PR #161, opened 2023-02-08
Use Makefile to generate manifest to use ledgerctl

mlafon-ledger:
## Description

Modify Makefile to create a manifest to sideload app with ledgerctl

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Makefile.rules_generic
+++ b/Makefile.rules_generic
@@ -71,6 +71,8 @@ DBG_TARGETS := debug/app.map debug/app.asm
 
 default: $(BIN_TARGETS) $(DBG_TARGETS)
 
+manifest: bin/app.toml
+
 $(OBJ_DIR)/%.o: %.c $(GLYPH_DESTC) prepare Makefile
 	@echo "[CC]   $@"
 	$(L)$(call cc_cmdline,$(INCLUDES_PATH), $(DEFINES),$<,$@)
@@ -159,3 +161,6 @@ SCAN_BUILD ?= scan-build
 SCAN_BUILD_OUTPUT ?= $(CURDIR)/output-scan-build
 scan-build:
 	$(SCAN_BUILD) --use-cc=$(CC) -analyze-headers -enable-checker security -enable-checker unix -enable-checker valist -o $(SCAN_BUILD_OUTPUT) $(MAKE) default
+
+bin/app.toml: bin/app.elf
+	python3 $(BOLOS_SDK)/create_manifest.py -output $@ $(APP_LOAD_PARAMS) --gif $(ICONNAME)
--- a/create_manifest.py
+++ b/create_manifest.py
@@ -1,0 +1,97 @@
+#!/usr/bin/env python3
+
+"""
+/*******************************************************************************
+*   Ledger SDK
+*   (c) 2022 Ledger
+*
+*  Licensed under the Apache License, Version 2.0 (the "License");
+*  you may not use this file except in compliance with the License.
+*  You may obtain a copy of the License at
+*
+*      http://www.apache.org/licenses/LICENSE-2.0
+*
+*  Unless required by applicable law or agreed to in writing, software
+*  distributed under the License is distributed on an "AS IS" BASIS,
+*  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+/*******************************************************************************
+"""
+
+import sys
+import argparse
+import tomli_w
+
+def auto_int(x):
+	return int(x, 0)
+
+# Same as ledgerblue/loadApp.py for compatibility
+def get_argparser():
+	parser = argparse.ArgumentParser(description="Load an app onto the device from a hex file.")
+	parser.add_argument("-output", help="File where the manifest is saved", required=True)
+	parser.add_argument("--targetId", help="The device's target ID (default is Ledger Blue)", type=auto_int, default=0x31000002)
+	parser.add_argument("--apiLevel", help="Set the API level of the SDK used to build the app", type=auto_int, default=0)
+	parser.add_argument("--fileName", help="The application hex file to be loaded onto the device", required=True)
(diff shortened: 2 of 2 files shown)


create_manifest.py
```diff
@@ -0,0 +1,97 @@
...
+*  distributed under the License is distributed on an "AS IS" BASIS,
+*  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+/*******************************************************************************
+"""
+
+import sys
+import argparse
+import tomli_w
```
xchapron-ledger, line 24: This is not a standard lib I guess? I mean not included by default like sys and argparse?

Makefile.rules_generic
```diff
@@ -68,8 +68,9 @@ prepare:
 
 BIN_TARGETS := bin/app.elf bin/app.apdu bin/app.sha256 bin/app.hex
 DBG_TARGETS := debug/app.map debug/app.asm
+MANIFEST := app.toml
 
-default: $(BIN_TARGETS) $(DBG_TARGETS)
+default: $(BIN_TARGETS) $(DBG_TARGETS) $(MANIFEST)
```
xchapron-ledger, line 73: Now that I've had a few issue with our app deployement piepeline, I'm not sure anymore we should add this to the default target. Even more if it needs custom python lib.

xchapron-ledger: Also, I'm not sure the MANIFEST vairable is really necessary.

Makefile.rules_generic
```diff
@@ -159,3 +160,6 @@ SCAN_BUILD ?= scan-build
 SCAN_BUILD_OUTPUT ?= $(CURDIR)/output-scan-build
 scan-build:
 	$(SCAN_BUILD) --use-cc=$(CC) -analyze-headers -enable-checker security -enable-checker unix -enable-checker valist -o $(SCAN_BUILD_OUTPUT) $(MAKE) default
+
+app.toml: bin/app.elf
```
xchapron-ledger, line 164: I'm not sure about the app.toml location, could it be somewhere else ? in bin ? in build/<target> ?

create_manifest.py
mlafon-ledger: Nop it is not a standard lib 
I made a PR in [ledger_app_builder](https://github.com/LedgerHQ/ledger-app-builder/pull/45) to have it

Makefile.rules_generic
mlafon-ledger: Ok

Makefile.rules_generic
mlafon-ledger: I can try to modify it to have it in bin or build

create_manifest.py
xchapron-ledger: Ok! But as this is not needed by many of external developers (that might not use ledger-app-builder) and our tools (that don't use ledger-app-builder) I suggest calling it only for a specific target, so that it doesn't create trouble when not necessary. I know I told you the opposite last time, sorry.

Makefile.rules_generic
mlafon-ledger: Done

create_manifest.py
```diff
@@ -0,0 +1,97 @@
...
+
+def auto_int(x):
+	return int(x, 0)
+
+def get_argparser():
+	parser = argparse.ArgumentParser(description="Load an app onto the device from a hex file.")
```
xchapron-ledger, line 30: Can you add a comment explaining this is taken as is from loadApp.py parser which can expect quite a lot of params...

create_manifest.py
```diff
@@ -0,0 +1,97 @@
...
+	parser.add_argument("--path_slip21", help="""A SLIP 21 path to which derivation is locked""", action='append')
+	parser.add_argument("--signature", help="A signature of the application (hex encoded)")
+	parser.add_argument("--signApp", help="Sign application with provided signPrivateKey", action='store_true')
+	parser.add_argument("--bootAddr", help="The application's boot address", type=auto_int)
+	parser.add_argument("--rootPrivateKey", help="""The Signer private key used to establish a Secure Channel (otherwise
+a random one will be generated)""")
+	parser.add_argument("--signPrivateKey", help="Set the private key used to sign the loaded app")
+	parser.add_argument("--apdu", help="Display APDU log", action='store_true')
+	parser.add_argument("--deployLegacy", help="Use legacy deployment API", action='store_true')
+	parser.add_argument("--apilevel", help="Use given API level when interacting with the device", type=auto_int, default=10)
```
xchapron-ledger, line 55: This parameter has been removed from loadApp.py and could therefore be removed here too.

create_manifest.py
xchapron-ledger: Maybe you could even import `get_argparser()` from `ledgerblue.loadApp`? It would make things more obvious. What do you think?

xchapron-ledger (commented):
Look good for me.
Now you need someone from the Os to review it too :)

create_manifest.py
mlafon-ledger: Which one ? 
I see all the these element in [ledgerctl/loadApp](https://github.com/LedgerHQ/blue-loader-python/blob/master/ledgerblue/loadApp.py) but not in the same order

xchapron-ledger: `apilevel` has been removed, but `apiLevel` has been kept (https://github.com/LedgerHQ/blue-loader-python/pull/100)

create_manifest.py
mlafon-ledger: So it means that users will have to use ledgerblue but the point is to use ledgerctl this is why I did not want to add ledgerblue dependancies. But I can add a comment and copy/paste the arguments in the same order that there are in ledgerblue if you want

create_manifest.py
mlafon-ledger: Thanks done

cbouvet-ledger (approved):

xchapron-ledger:
I'm wondering, couldn't we use the already existing `--offline` parameter from `ledgerblue/loadµApp.py`. It's purpose is to generate an `app.apdu` which should contains everything. But maybe ledgerctl doesn't support loading app from app.apdu file?

PR #184, opened 2023-02-21
Remove clang warning

dmorais-ledger:
This PR removes some clang warnings.


lib_bagl/src/bagl.c
```diff
@@ -374,42 +374,42 @@ int bagl_draw_string(unsigned short font_id, unsigned int fgcolor, unsigned int
 
       // 6 bytes UTF-8, Unicode 0x400 0000 to 0x7FFF FFFF
       /*if (ch >= 0xFC && text_length >= 5) {
-        unicode = (ch - 0xFC) << 30;
```
xchapron-ledger, line 377: Is there a reason to keep this code which is commented out?

dmorais-ledger: It was to make things easier if one day we have UTF-8 codes using more than 4 bytes, but now it is in the history, so no problem.

cbouvet-ledger (approved):

PR #190, opened 2023-02-23
Remove DES-based algorithms

yhql:
## Description

The DES/3DES algorithms usage should be deprecated. This PR removes all files and comments associated with it.
Looking for any use of it in our apps shows no result, it should be safe to do so currently.

## Changes include

## Breaking changes

APIs relating to DES do not exist anymore.


lib_cxng/cx.export
srasoamiaramanana-ledger, line 1: I think you shouldn't remove this file but only remove the DES functions and then generate the content of `cx_stubs.h`, `cx_exported_functions.c` and `cx_stubs.S` with `cxsdk.py`. This way, the values in `cx_stubs.h` are realigned and that's easier to export new functions.

srasoamiaramanana-ledger (changes requested):

srasoamiaramanana-ledger (approved):

PR #213, opened 2023-03-17
Add a new error code

srasoamiaramanana-ledger:


srasoamiaramanana-ledger:
This is linked to a PR which is paused. I guess, it is not an issue to merge this as it is just an error code which can be used anyway.

PR #230, opened 2023-03-24
put back blake2 and ripemd hash algo

tjoly-ledger:


lib_cxng/include/lcx_groestl.h
```diff
@@ -85,6 +85,8 @@ size_t cx_groestl_get_output_size(const cx_groestl_t *ctx);
  *
  * @param[in]  size Length of the digest.
  *
+ * @return          GROESTL identifier.
+ *
```
xchapron-ledger, line 89: this function is void, so it don't return anything?

xchapron-ledger (commented):
Is it possible to use perfect `git revert` of merged commits? So that it somehow easier to review?

lib_cxng/include/lcx_groestl.h
tjoly-ledger: It is from the revert, I prefer to no update anything.

PR #239, opened 2023-03-28
Makefile: Fix build of app that doesn't use default target

xchapron-ledger:
## Description

This fix a regression from 01d5cd4f9eb69ca6071736af7ebc3d832da1d8bd

> Some app use `bin/app.elf` as default makefile target instead of `default`.
> Therefore we should keep for now the `prepare` dependency on next target
> in order not to break these apps build...
> At some point, this should be changed in the apps and removed from the SDK.

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Makefile.rules_generic
+++ b/Makefile.rules_generic
@@ -79,7 +79,11 @@ default: prebuild
 	$(MAKE) glyphs_gen
 	$(MAKE) targets_gen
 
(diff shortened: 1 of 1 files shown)


sgliner-ledger (approved):

tjoly-ledger (approved):
LGTM

PR #250, opened 2023-04-04
Nbgl fixes for fonts and navigation

nroggeman-ledger:
## Description

The goal is to fix a minor issue on navigation bar, which is causing a refresh even when navigation arrow is inactive.
It also includes improvement on nbgl_fonts management (factorization of character retrieval)

## Changes include

- [*] Bugfix (non-breaking change that solves an issue)
- [ ] Tests
- [ ] Documentation


cbouvet-ledger:
a cherry pick en 7

PR #290, opened 2023-04-26
Update NBGL use case wording to be compliant with the Figma reference layouts

tdejoigny-ledger:
## Description

Update NBGL use case wording to be compliant with the Figma layouts

## Changes include

- [ ] Tests
- [ ] Documentation


xchapron-ledger (approved):

tjulien-ledger (approved):
Thank you for applying the changes, these are indeed inline with recommendations!

PR #296, opened 2023-05-05
Add __attribute__((noreturn)) for os_lib_end SYSCALL

fbeutin-ledger:
## Description

An attribute was missing, it triggered false positive warnings at compilation
## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/include/os_lib.h
+++ b/include/os_lib.h
@@ -12,4 +12,8 @@
  * call_parameters[2+] = called function parameters
  */
 SYSCALL void os_lib_call(unsigned int *call_parameters PLENGTH(3*sizeof(unsigned int)));
+#ifdef HAVE_BOLOS
 SYSCALL void os_lib_end(void);
+#else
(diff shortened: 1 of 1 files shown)


include/os_lib.h
```diff
@@ -12,4 +12,4 @@
  * call_parameters[2+] = called function parameters
  */
 SYSCALL void os_lib_call(unsigned int *call_parameters PLENGTH(3*sizeof(unsigned int)));
-SYSCALL void os_lib_end(void);
+SYSCALL void __attribute__((noreturn)) os_lib_end(void);
```
xchapron-ledger, line 15: Maybe this should be under HAVE_BOLOS ifdef has here: https://github.com/LedgerHQ/ledger-secure-sdk/blob/master/include/os_task.h#L31-L35

fbeutin-ledger: Done

xchapron-ledger (approved):

PR #343, opened 2023-06-08
nbgl: handle very long app names

sgliner-ledger:
## Description

Very long appnames resulted in a truncated string in the home screen.
This solve the issue by defining a default string in case the app fails at providing a custom string.

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/lib_nbgl/include/nbgl_use_case.h
+++ b/lib_nbgl/include/nbgl_use_case.h
@@ -54,6 +54,22 @@ extern "C" {
  */
 #define TAG_VALUE_AREA_HEIGHT   400
 
+/**
+ *  @brief Default strings used in the Home tagline
+ */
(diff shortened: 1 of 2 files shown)


xchapron-ledger (approved):

lib_nbgl/include/nbgl_use_case.h
```diff
@@ -54,6 +54,22 @@ extern "C" {
...
+/**
+ *  @brief Default strings used in the Home tagline
+ */
+#define TAGLINE_PART1 "This app enables signing\ntransactions on the"
+#define TAGLINE_PART2 "network."
+
+/**
+ *  @brief Length of buffer used for the default Home tagline
+ */
+#define APP_DESCRIPTION_MAX_LEN 72
```
tdejoigny-ledger, line 66: should we increase a bit this value to "support" Binance Smart Chain app ?

PR #346, opened 2023-06-13
Add API_LEVEL_5 JSON files for Nano fonts

dmorais-ledger:
## Description
Contains the .json files with font data for Nano X & Nano S+
(those files are JSON versions of the .inc files, to be able to use them with Python to fix speculos OCR troubles)


lib_bagl/src/bagl_font_open_sans_extrabold_11px.inc
```diff
@@ -185,7 +183,8 @@ const unsigned char bitmapOPEN_SANS_EXTRABOLD_11PX[] = {
 //ascii 0x0078
   0x77, 0x1B, 0x87, 0x63, 0xBB, 0x03,
 //ascii 0x0079
-  0x63, 0x9B, 0x8D, 0xC6, 0xE1, 0x30, 0x18, 0x06,
+  0x63, 0x36, 0x36, 0x34, 0x1C, 0x1C, 0x1C, 0x0C,
+  0x0E,
```
xchapron-ledger, line 187: Is this intended? If I understand correctly this will change the apperance of the "y" letter?

dmorais-ledger: I can't generate .json files without generating corresponding .inc files at the same time, and as you noticed, some changes have been made on some letters.
(It's supposed to be an improvement, not a regression)

Generating an app without using those .inc files is not mandatory, but it will result in OCR failing to recognize the couple of letters which are different.

An other solution would be to write a **inc2json.py** script, but this is a lot of work for just _one shot_...

dmorais-ledger: As I told you on Slack, I was able to use the pictures corresponding to API_LEVEL_5 tag and the generated .inc files was identical => this PR will now just include the .json files.

xchapron-ledger: Did you pushed it? I still see changes in `.inc` files.

xchapron-ledger:
@dmorais-ledger Can we indent the JSON files?
What do you think of checking that they indeed correspond to the `.inc` files with a CI check maybe?

dmorais-ledger:
> @dmorais-ledger Can we indent the JSON files? 

Those JSON files are not intended to be read by human, but its done.

> What do you think of checking that they indeed correspond to the `.inc` files with a CI check maybe?

The .inc & .json files are generated at the same time by the **ttf2inc.py** script, which is now part of the SDK.
I don't see why anyone would modify one of these files.
If someone really want to change a glyph, the simplest way to do it is to change the corresponding picture.
(they are located in **lib_bagl/fonts**, in latest SDKs)

xchapron-ledger (approved):

PR #347, opened 2023-06-15
Generate all JSON files - useful for speculos OCR

dmorais-ledger:
## Description

This PR includes latest JSON files containing font information, used by speculos OCR.

Those files will automatically be generated by Bolos, once this PR will be validated.


xchapron-ledger:
@dmorais-ledger we need to find a way to exclude the false positive detection of spellcheck...

PR #350, opened 2023-06-16
STAX Touch debug data syscalls factorization

PhilippeBonnaz:
## Description

Syscalls with ID SYSCALL_touch_read_sensi_ID and SYSCALL_touch_read_diff_data_ID are merged into one syscall with ID SYSCALL_touch_debug_ID
A second parameter has been added to the merged syscall to allow "dispatching" the two previous syscalls
So depending on this parameter value, this new syscall can:
- fetch touch raw data
- fetch touch diff data
- set touch panel in "read coordinates mode" (used by a caller script, to setup STAX in "standard mode" (ie read coordinates) after test

Breaking API change with SDK functions renaming (touch debug functions only used by STAX touch test apps)

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/os_io_seproxyhal.c
+++ b/src/os_io_seproxyhal.c
@@ -501,12 +501,34 @@ void io_seproxyhal_nfc_init(ndef_struct_t *ndef_message, bool async, bool forceI
 
 #ifdef HAVE_SE_TOUCH
 #ifdef HAVE_TOUCH_READ_DEBUG_DATA_SYSCALL
-void io_seproxyhal_touch_read_sensi(uint8_t * sensi_data) {
-  touch_read_sensitivity(sensi_data);
+/**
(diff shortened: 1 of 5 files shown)


src/os_io_seproxyhal.c
```diff
@@ -508,6 +508,10 @@ void io_seproxyhal_touch_read_sensi(uint8_t * sensi_data) {
 void io_seproxyhal_touch_read_diff_data(uint8_t *diff_data) {
   touch_read_diff_data(diff_data);
 }
+
+void io_seproxyhal_touch_read_coordinates_mode(void) {
```
cbouvet-ledger, line 512: why is that function needed? that's not even an "io" function

abonnaudet-ledger (approved):

PR #354, opened 2023-06-21
lib_standard_app: main.c: Improvements

xchapron-ledger:
## Description

Previous behavior was to catch the Exception and somehow restart the app. This was somehow hiding the error as the user would not see the app crash. As Exception should not be raised in nominal situation or should be caught before, we changed the behavior to directly exit the app.

This also have the benefit of avoiding issue where the whole app BSS should be reset to have a sane behavior when the app was restarted.

Linked to https://github.com/LedgerHQ/app-boilerplate/pull/85

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/lib_standard_app/main.c
+++ b/lib_standard_app/main.c
@@ -26,15 +26,8 @@ bolos_ux_params_t G_ux_params;
 /**
  * Exit the application and go back to the dashboard.
  */
-WEAK void app_exit() {
-    BEGIN_TRY_L(exit) {
-        TRY_L(exit) {
-            os_sched_exit(-1);
-        }
-        FINALLY_L(exit) {
-        }
-    }
-    END_TRY_L(exit);
+WEAK void __attribute__((noreturn)) app_exit(void) {
+    os_sched_exit(-1);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


lib_standard_app/main.c
```diff
@@ -45,40 +38,35 @@ WEAK __attribute__((section(".boot"))) int main() {
...
-            CATCH_ALL {
-                CLOSE_TRY;
-                break;
-            }
-            FINALLY {
-            }
+
+            app_main();
+        }
+        CATCH_OTHER(e) {
```
sgliner-ledger, line 61: What is the difference between `CATCH_OTHER` and `CATCH_ALL` ?

sgliner-ledger (approved):

lib_standard_app/main.c
xchapron-ledger: Discussed in direct

lib_standard_app/main.c
```diff
@@ -38,40 +38,34 @@ WEAK __attribute__((section(".boot"))) int main() {
 
     os_boot();
 
-    for (;;) {
-        // Initialize the UX system
-        UX_INIT();
+    BEGIN_TRY {
+        TRY {
+            UX_INIT();
```
sgliner-ledger, line 43: Do we remove this ?

xchapron-ledger: It's still used for NBGL, and can be a placeholder for futur changes?
So I'd say we should keep it.

PR #386, opened 2023-07-28
Tell codespell to skip JSON files in fonts directories

dmorais-ledger:
Tell codespell to skip JSON files in fonts directories


xchapron-ledger:
done in https://github.com/LedgerHQ/ledger-secure-sdk/pull/387

PR #398, opened 2023-08-02
Unify linkscript for all targets

yhql:
## Description

We currently use one linker script per target that embeds the memory definitions and the placement of code sections in memory. The placement part is pretty similar across devices and has no reason to be kept duplicated.
This PR splits things into per-target memory definitions and a single `script.ld` that places code sections, and can be updated for all devices at once when we need it to.
Perhaps even the memory definitions could be merged together also (in another PR).

## Changes include

(cleanup/refactor)


xchapron-ledger:
This is breaking app-bitcoin-new build on nanos, so probably something doesn't works as expected.
We probably should check that this doesn't change app binaries.

bigspider:
I suppose it's because the `G_cx` symbol is no longer allocated in the CXRAM section?
Note that I use the CXRAM section explicitly on NanoS, when I need memory but I'm no using any crypto functions that might touch CXRAM... 😇

yhql:
Yes, I forgot the placement of G_cx. Moved it to the Nano S's layout only.

target/script.ld
tdejoigny-ledger, line 1: Not sure if it is really used but the "debug" part for each device has been dropped

tdejoigny-ledger, line 1: why you removed some comments? (some of them seem to be useful)

yhql: I found those to be either invalid (or not valid anymore) or confusing

PR #401, opened 2023-08-08
Keep only sdk errors in errors.h

srasoamiaramanana-ledger:
## Description

Keep only sdk errors in errors.h.

## Changes include

- [ ] Tests
- [ ] Documentation


srasoamiaramanana-ledger:
Change `ERR_APD_LEN` to `ERR_APP_APD_LEN` in app-mimblewimblecoin: https://github.com/LedgerHQ/app-mimblewimblecoin/blob/master/src/continue_decrypting_slate.c#L75

include/errors.h
```diff
@@ -188,601 +125,22 @@ enum generic_identifiers {
...
-#define SWO_APD_STA_35                    (ERR_APD_STA + ERR_GEN_ID_35)    // 0x6635
-#define SWO_APD_STA_36                    (ERR_APD_STA + ERR_GEN_ID_36)    // 0x6636
-#define SWO_APD_STA_37                    (ERR_APD_STA + ERR_GEN_ID_37)    // 0x6637
-#define SWO_APD_STA_38                    (ERR_APD_STA + ERR_GEN_ID_38)    // 0x6638
-#define SWO_APD_STA_39                    (ERR_APD_STA + ERR_GEN_ID_39)    // 0x6639
-#define SWO_APD_STA_3A                    (ERR_APD_STA + ERR_GEN_ID_3A)    // 0x663A
-#define SWO_APD_STA_3B                    (ERR_APD_STA + ERR_GEN_ID_3B)    // 0x663B
-#define SWO_APD_STA_3C                    (ERR_APD_STA + ERR_GEN_ID_3C)    // 0x663C
-#define SWO_APD_STA_3D                    (ERR_APD_STA + ERR_GEN_ID_3D)    // 0x663D
+#ifndef SWO_SEC_PIN_15
```
xchapron-ledger, line 135: Do you need the `#ifndef`? As the value is the same, there is probably no issue with redefining the same value?

include/errors.h
```diff
@@ -188,601 +125,22 @@ enum generic_identifiers {
...
- * Values mismatch in functions (VAL).
- */
-#define ERR_CRY_LEN                       (ERR_CRY_RANGE + ERR_GEN_SUB_01)
-#define ERR_CRY_VAL                       (ERR_CRY_RANGE + ERR_GEN_SUB_02)
+#define ERR_APP_APD_CLA                   (ERR_APP_APD_RANGE + ERR_GEN_SUB_0E) // 0x6E00
+#define ERR_APP_APD_INS                   (ERR_APP_APD_RANGE + ERR_GEN_SUB_0D) // 0x6D00
+#define ERR_APP_APD_HDR                   (ERR_APP_APD_RANGE + ERR_GEN_SUB_05) // 0x6500
+#define ERR_APP_APD_STA                   (ERR_APP_APD_RANGE + ERR_GEN_SUB_06) // 0x6600
+#define ERR_APP_APD_LEN                   (ERR_APP_APD_RANGE + ERR_GEN_SUB_07) // 0x6700
+#define ERR_APP_APD_DAT                   (ERR_APP_APD_RANGE + ERR_GEN_SUB_08) // 0x6800
```
xchapron-ledger, line 133: Why do you need to add _APP in the name?
This introduce a breaking change that doesn't seems necessary? Though this doesn't break many apps it seems.

include/errors.h
```diff
@@ -53,6 +28,14 @@
 #define ERR_APP_RANGE_03                  0xD000
 #define ERR_APP_RANGE_04                  0xE000
 
+/**
+ * I/O error codes range.
+ *
+ */
+#define ERR_IOL_RANGE                     0x1000
+
+#define ERR_APP_APD_RANGE                 0x6000
```
xchapron-ledger, line 37: Same question about the renaming. Couldn't we keep the same name?

include/errors.h
srasoamiaramanana-ledger: `ERR_APD_RANGE` is used to define the OS error codes. Keeping the name makes a redefinition which is not allowed.

xchapron-ledger (approved):

PR #402, opened 2023-08-08
hkdf_expand info arg can be const

NicsTr:


NicsTr:
## Description

hkdf_expand info arg is set to const

## Changes include

- [ ] Tests
- [ ] Documentation

yhql (approved):

PR #403, opened 2023-08-09
[BLE] enable use of local APDU buffer

yhql:
## Description

This PR primarily submits changes to match its use in the Rust SDK. 

- The use of `HAVE_LOCAL_APDU_BUFFER` means `G_io_apdu_buffer` does not exist anymore and a way to know where to copy received commands is put in place. An additional pointer is added to the main BLE context, and all copies to the global apdu buffer are replaced with copies to this new pointer.
Its value and associated length can be set externally through a new function. The use of `G_io_seproxyhal_spi_buffer` is also replaced with locals where possible to ease tracking data moves.

- To make sure bindings to the C APIs are correct, the use of `const` was enforced were necessary, sometimes requiring modifying the code a bit (see the `DEF1` channel ID copy that was moved).

- Functions that enable BLE pairing while in an app were gated so that they can be deactivated, which allows not having to link in `snprintf` and other functions from another module (ux)

## Changes include


Makefile.defines
```diff
@@ -212,6 +212,7 @@ DEFINES       += HAVE_MCU_SERIAL_STORAGE
 DEFINES       += HAVE_FONTS
 # already defined within apps
 # DEFINES       += HAVE_BLE
+DEFINES       += HAVE_INAPP_BLE_PAIRING
```
xchapron-ledger, line 215: Should be used on Stax too I think.

yhql: true, thanks

yhql:
Tested as applied on API_LEVEL_5 on a Nano X 2.2.2, on `rust-app` ("menu" and a "sign" command), and on the current `app-boilerplate` with a "get name and version" command.

src/os_io_seproxyhal.c
```diff
@@ -1004,8 +1004,9 @@ void io_seproxyhal_power_off(bool criticalBattery)
         (char) criticalBattery,
     };
     io_seproxyhal_spi_send(seph_io_device_off, sizeof(seph_io_device_off));
-    for (;;)
+    for (;;) {
         ;
+    }
```
xchapron-ledger, line 1009: Can tou drop this change that isn't link with this PR and might make the cherry-pick process more difficult.

src/os_io_seproxyhal.c
```diff
@@ -1016,8 +1017,9 @@ static const unsigned char seph_io_se_reset[] = {
 void io_seproxyhal_se_reset(void)
 {
     io_seproxyhal_spi_send(seph_io_se_reset, sizeof(seph_io_se_reset));
-    for (;;)
+    for (;;) {
```
xchapron-ledger, line 1020: Same as https://github.com/LedgerHQ/ledger-secure-sdk/pull/403/files#r1325473922

sportron-ledger:
Hello,
What is the purpose of the HAVE_INAPP_BLE_PAIRING define?

yhql:
@sportron-ledger it allows disabling parts of the code that require specific dependencies on `printf` and `ux` machinery that is not present (or not easy to integrate) in the Rust SDK.

src/os_io_seproxyhal.c
yhql: good catch thanks

PR #424, opened 2023-09-12
Add io_send_response_buffers() to lib_standard_app

elric1:
## Description

We add a function `io_send_response_buffers()` which takes an array of buffers rather than just a single buffer. We also rewrite the existing `io_send_response_buffer()` in terms of the new function (trivially).

## Changes include

- [ ] Tests
- [ ] Documentation


lib_standard_app/io.c
```diff
@@ -202,6 +208,10 @@ WEAK int io_send_response_buffer(const buffer_t *rdata, uint16_t sw) {
     return ret;
 }
 
+WEAK int io_send_response_buffer(const buffer_t *rdata, uint16_t sw) {
+    return io_send_response_buffers(rdata, 1, sw);
+}
+
 WEAK int io_send_sw(uint16_t sw) {
     return io_send_response_buffer(NULL, sw);
 }
```
xchapron-ledger, line 217: I'm wondering if we should inline those and put them on io.h to save a few bytes of flash...

xchapron-ledger: maybe this overkilled

lib_standard_app/io.c
```diff
@@ -153,19 +153,25 @@ WEAK int io_send_response_pointer(const uint8_t *ptr, size_t size, uint16_t sw)
...
+        for (size_t i=0; i < count; i++) {
+            const buffer_t *rdata = &rdatalist[i];
+
+            if (G_output_len + rdata->size - rdata->offset > IO_APDU_BUFFER_SIZE - 2 ||
+                !buffer_copy(rdata, G_io_apdu_buffer + G_output_len,
+                             sizeof(G_io_apdu_buffer) - G_output_len)) {
+                return io_send_sw(SW_WRONG_RESPONSE_LENGTH);
+            }
+            G_output_len += rdata->size - rdata->offset;
+            PRINTF("<= SW=%04X | (%u) RData=%.*H\n", sw, i, rdata->size, rdata->ptr);
```
xchapron-ledger, line 170: Should be moved out the for loop.

lib_standard_app/io.c
xchapron-ledger: After discussion, let do it, we have already do simiar stuff on crypto_helper.h

lib_standard_app/io.c
```diff
@@ -149,20 +149,23 @@ WEAK int io_recv_command() {
...
+        for (size_t i=0; i < count; i++) {
+            const buffer_t *rdata = &rdatalist[i];
+
+            if (G_output_len + rdata->size - rdata->offset > IO_APDU_BUFFER_SIZE - 2 ||
+                !buffer_copy(rdata, G_io_apdu_buffer + G_output_len,
+                             sizeof(G_io_apdu_buffer) - G_output_len)) {
+                return io_send_sw(SW_WRONG_RESPONSE_LENGTH);
+            }
+            G_output_len += rdata->size - rdata->offset;
+            PRINTF("<= FRAG (%u/%u) RData=%.*H\n", i+1, count, rdata->size, rdata->ptr);
```
xchapron-ledger, line 166: Can you print this only if `count > 1` ? So that log for for response sending without fragment doesn't have frag message.

elric1: Yes, I just updated it.

xchapron-ledger:
@elric1 Could you please fix the lint failure: https://github.com/LedgerHQ/ledger-secure-sdk/actions/runs/6186093143/job/16831576912?pr=424

lib_standard_app/io.h
```diff
@@ -96,4 +119,7 @@ WEAK int io_send_response_buffer(const buffer_t *rdata, uint16_t sw);
  * @return zero or positive integer if success, -1 otherwise.
  *
  */
-WEAK int io_send_sw(uint16_t sw);
+static inline int io_send_sw(uint16_t sw)
+{
+    return io_send_response_buffer(NULL, sw);
```
xchapron-ledger, line 124:
```suggestion
    return io_send_response_buffers(NULL, 0, sw);
```
Else you are going to deference NULL

xchapron-ledger (changes requested):

xchapron-ledger (approved):

lib_standard_app/io.c
```diff
@@ -153,22 +153,27 @@ WEAK int io_recv_command()
...
+        for (size_t i = 0; i < count; i++) {
+            const buffer_t *rdata = &rdatalist[i];
+
+            if (G_output_len + rdata->size - rdata->offset > IO_APDU_BUFFER_SIZE - 2
+                || !buffer_copy(rdata,
+                                G_io_apdu_buffer + G_output_len,
+                                sizeof(G_io_apdu_buffer) - G_output_len)) {
+                return io_send_sw(SW_WRONG_RESPONSE_LENGTH);
+            }
+            G_output_len += rdata->size - rdata->offset;
```
yhql, line 171: Could add an overflow check here to avoid writing the status word at a bad place, or sending more than G_io_apdu_buffer

lib_standard_app/io.c
```diff
@@ -153,22 +153,27 @@ WEAK int io_recv_command()
...
-            return io_send_sw(SW_WRONG_RESPONSE_LENGTH);
+    if (rdatalist && count > 0) {
+        G_output_len = 0;
+        for (size_t i = 0; i < count; i++) {
+            const buffer_t *rdata = &rdatalist[i];
+
+            if (G_output_len + rdata->size - rdata->offset > IO_APDU_BUFFER_SIZE - 2
+                || !buffer_copy(rdata,
+                                G_io_apdu_buffer + G_output_len,
+                                sizeof(G_io_apdu_buffer) - G_output_len)) {
```
xchapron-ledger, line 168: This should be enough and less confusing, buffer_copy already check for reception buffer size.
```suggestion
            if (!buffer_copy(rdata,
                             G_io_apdu_buffer + G_output_len,
                             sizeof(G_io_apdu_buffer) - G_output_len - 2)) {
```
@yhql what do you think?

elric1: I don't think that this change will work as intended because the final argument to buffer_copy() is size_t. size_t is an unsigned type which means that without an appropriate guard, if it underflows then it will become (on most implementations) a very large positive number and thus the check will not prevent a buffer overflow.

xchapron-ledger: I can't see how this could happen?
I we expand the buffer_copy check it gives us something like:

```
if (buffer->size - buffer->offset > sizeof(G_io_apdu_buffer) - G_output_len - 2) {
    return false;
}
```

if `buffer->size < buffer->offset` it will for sure underflow, but then it will be > `sizeof(G_io_apdu_buffer) - G_output_len - 2` which by construction can't underflow as `G_output_len` starts from `0` and is increased only if the condition above is respected.
Considering that `sizeof(G_io_apdu_buffer) - G_output_len - 2` won't underflow by construction, then in fact `buffer_copy` check is identic as the check I'm proposing to remove (`G_output_len + rdata->size - rdata->offset > IO_APDU_BUFFER_SIZE - 2`).

So I don't see how bad things could occurs.
Am I missing something here?

elric1: When you call:
```
       buffer_copy(rdata,
                             G_io_apdu_buffer + G_output_len,
                             sizeof(G_io_apdu_buffer) - G_output_len - 2)
```
Let's say that G_output_len = sizeof(G_io_apdu_buffer) + 3. That argument is evaluated before being passed to the function. Using signed arithmetic, that would be -1. But, the variable is not signed and so it will end up being 0xffffffff which is coincidentally MAX_INT. When we get to the check in buffer_copy(), we see:
```
    if (buffer->size - buffer->offset > out_len)
```
our out_len in this example is 0xffffffff which can _never_ be true because that's MAX_INT.
Please keep in mind that I am assuming two compliment arithmetic and all that in this example. According to the C standard, the behaviour is undefined meaning that anything could happen.

xchapron-ledger: > Let's say that `G_output_len = sizeof(G_io_apdu_buffer) + 3`
This is not possible by construction, `G_output_len` max value is `sizeof(G_io_apdu_buffer) - 2`.
Else we have other issues, e.g.g already a bffer overflow on the previous buffer copy.

xchapron-ledger: @elric1 An up to date comment is always good!

elric1: I force pushed a new version. I also dedup `G_output_len = 0;` because it doesn't need to be set in both branches of the conditional when it is perfectly acceptable to set it at the beginning.

lib_standard_app/io.c
```diff
@@ -153,25 +153,28 @@ WEAK int io_recv_command()
...
-            !buffer_copy(rdata, G_io_apdu_buffer, sizeof(G_io_apdu_buffer))) {
-            return io_send_sw(SW_WRONG_RESPONSE_LENGTH);
+    G_output_len = 0;
+    if (rdatalist && count > 0) {
+        for (size_t i = 0; i < count; i++) {
+            const buffer_t *rdata = &rdatalist[i];
+
+            if (!buffer_copy(rdata,
+                             G_io_apdu_buffer + G_output_len,
+                             sizeof(G_io_apdu_buffer) - G_output_len)) {
```
xchapron-ledger, line 167:
```suggestion
                             sizeof(G_io_apdu_buffer) - G_output_len - 2)) {
```

xchapron-ledger (approved):

PR #432, opened 2023-09-19
Deprecate "check_api_level" function

tdejoigny-ledger:
## Description

Deprecate `check_api_level` function

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/include/os_apilevel.h
+++ b/include/os_apilevel.h
@@ -19,7 +19,15 @@
 #ifndef OS_APILEVEL_H
 #define OS_APILEVEL_H
 
+#ifndef HAVE_BOLOS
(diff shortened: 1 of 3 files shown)


xchapron-ledger (commented):
I think we could maybe also deprecate the defines from `include/os_apilevel.h` by using something like: 
```
#define CX_APILEVEL        _Pragma ("GCC warning \"Deprecated constant!\"") 12
#define CX_COMPAT_APILEVEL _Pragma ("GCC warning \"Deprecated constant!\"") 12
```

xchapron-ledger (approved):

PR #443, opened 2023-10-03
Fix #1364: Error codes not systematically checked in the implementation

dmorais-ledger:
Fix [issue #1364](https://git.donjon.ledger.fr/gitea/ledger-core/bolos-ng/issues/1364) about **Error codes not systematically checked in the implementation**

WARNING: Should be merged at the same time than corresponding PR on TTYT & bolos-ng:
https://git.donjon.ledger.fr/gitea/CI/toutouyoutest/pulls/192
https://git.donjon.ledger.fr/gitea/ledger-core/bolos-ng/pulls/2098

### DIFF
--- a/lib_cxng/include/lcx_ripemd160.h
+++ b/lib_cxng/include/lcx_ripemd160.h
@@ -59,7 +59,7 @@ typedef struct cx_ripemd160_s cx_ripemd160_t;
  * @return          Error code:
  *                  - CX_OK on success
  */
-cx_err_t cx_ripemd160_init_no_throw(cx_ripemd160_t *hash);
+WARN_UNUSED_RESULT cx_err_t cx_ripemd160_init_no_throw(cx_ripemd160_t *hash);
 
 /**
  * @brief   Initializes a RIPEMD-160 context.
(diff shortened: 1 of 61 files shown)


lib_cxng/include/lcx_ripemd160.h
```diff
@@ -71,7 +71,10 @@ cx_err_t cx_ripemd160_init_no_throw(cx_ripemd160_t *hash);
  */
 static inline int cx_ripemd160_init(cx_ripemd160_t *hash)
```
xchapron-ledger, line 72: This one should be `DEPRECATED` no?

dmorais-ledger: You're certainly right, but its like that on master and its not the topic of this PR, so I leave this for some volunteer ;-)
(and I just checked, it is not called today - but its not a reason to not fix it and check for other prototypes in the same case)

srasoamiaramanana-ledger (approved):

srasoamiaramanana-ledger:
Let's keep in mind that these changes might cause a bunch of warnings when building apps.

PR #451, opened 2023-10-16
Fix the processing of apdus across multiple sources

yrichard-ledger:
## Description

This PR fixes edge cases when several apdus are sent to the device from different media.

The expected behaviour when two apdus are received in parallel is now:
    * The first apdu processing is not impacted by the second one
    * The second apdu is not processed and the device replies with an error SWO_IOL_STA_02 (0x1302)

This PR also adds a set of functions so that an application can restrict the apdu source to a single media.
This is intended to lock the media during a complex flow of apdus. In particular during an app installation 
we can now reject automatically apdus from other medias

## Changes include

- [*] Bugfix (non-breaking change that solves an issue)
- [ ] Tests
- [ ] Documentation


tdejoigny-ledger:
@yrichard-ledger did you manage the case where we receive a new APDU (via BLE) when still processing the previous APDU (no answer provided) received also via BLE ?

yrichard-ledger:
> @yrichard-ledger did you manage the case where we receive a new APDU (via BLE) when still processing the previous APDU (no answer provided) received also via BLE ?

Yes, whatever the media, if we receive a new APDU while the previous apdu is still processing the SDK will reply with SWO_IOL_STA_02 (0x1302)

tdejoigny-ledger:
@yrichard-ledger as discussed, could we revisit this topic? I realize that we had already encountered some problematic cases with the BT [here](https://github.com/LedgerHQ/ledger-secure-sdk/pull/451#issuecomment-1931764445).
cc: @nroggeman-ledger

PR #473, opened 2023-11-27
Ledger assert

sgliner-ledger:
## Description

Add `legder-assert` and related defines to enable feature and display of error code on device screen.

## Changes include

- [ ] Tests
- [ ] Documentation

## Breaking changes

Replace `HAVE_DEBUG_THROWS` with `HAVE_LEDGER_ASSERT_DISPLAY`


src/os.c
```diff
@@ -86,22 +86,21 @@ char os_secure_memcmp(const void *src1, const void *src2, size_t length)
 
 #ifndef HAVE_BOLOS
 #define MAIN_LINKER_SCRIPT_LOCATION 0xC0DE0000
+#include "lib_standard_app/debug.h"
```
xchapron-ledger, line 89: Why do you need this here?

src/os.c
```diff
@@ -86,22 +86,21 @@ char os_secure_memcmp(const void *src1, const void *src2, size_t length)
...
 int main(void);
 
-// This function can be used to declare a callback to THROW in the application
-__attribute((weak)) void app_throw_info(unsigned int exception, unsigned int lr_val)
```
xchapron-ledger, line 92: How does that work for app that doesn't build standard app c files?

lib_standard_app/debug.c
```diff
@@ -59,13 +60,42 @@ WEAK void __attribute__((noreturn)) debug_display_throw_error(int exception)
...
     // Block until the user approve and the app is quit
     while (1) {
         io_seproxyhal_io_heartbeat();
     }
 }
+#endif
+
+#ifdef HAVE_LEDGER_ASSERT
+#define ASSERT_ERROR_CODE 0xBAAD
+void ledger_assert(char *message)
```
xchapron-ledger, line 75: With a function like that, the message string will always be embedded in the app binary, even when unused (when PRINTF are disabled).

I think we should go for an ASSERT made of a macro calling different functions in the lines of what's done here: https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/sys/__assert.h

- the ASSERT macro check for a result, so that it can be called unconditionally in the app code.
- the ASSERT macro can display a message or not, but doesn't hold it in the app binary if it's not being displayed
- the ASSERT macro can display file name and line, but this should be disabled by default as this use a lot of flash
- the ASSERT macro can access both the PC and the LR (instead of just the PC which is called LR here as it's a macro and not a function).

src/os.c
xchapron-ledger: I don't think this will work on all apps.
they will have the .h, but not the .c if not using standard_app, hence they can't use `HAVE_DEBUG_THROWS`, which is not really explicit?

xchapron-ledger (commented):
I'm wondering if we should put this out of standard_app? So that it is used more simply on app not using standard_app?

Makefile.standard_app
```diff
@@ -60,6 +60,14 @@ else
         DEFINES += PRINTF\(...\)=
 endif
 
+#####################################################################
+#                          LEDGER_ASSERT                            #
+#####################################################################
+ifneq ($(DISABLE_LEDGER_ASSERT), 1)
+	DEFINES += HAVE_LEDGER_ASSERT_DISPLAY
```
xchapron-ledger, line 67: Are we sure we want this by default? Exposing end user of weird info (rather than an app exit without info, maybe we need a product on this?)

Makefile.standard_app
```diff
@@ -60,6 +60,14 @@ else
         DEFINES += PRINTF\(...\)=
 endif
 
+#####################################################################
+#                          LEDGER_ASSERT                            #
+#####################################################################
```
xchapron-ledger, line 65: Maybe when DEBUG is on, we should enable a bigger level of displayed info? Like the file and line?

lib_standard_app/debug.h
```diff
@@ -1,3 +1,168 @@
...
+#ifdef LEDGER_ASSERT_CONFIG_LR_AND_PC_INFO
+#define LEDGER_ASSERT_LR_AND_PC()                            \
+    do {                                                     \
+        int _lr_address = 0;                                 \
+        int _pc_address = 0;                                 \
+                                                             \
+        __asm volatile("mov %0, lr" : "=r"(_lr_address));    \
+        __asm volatile("mov %0, pc" : "=r"(_pc_address));    \
+        _lr_address = compute_address_location(_lr_address); \
+        _pc_address = compute_address_location(_pc_address); \
```
xchapron-ledger, line 119: To save memory on apps, and as this will probably be the default mode, we could move the `compute_address_location` inside `print_lr_and_pc` and `assert_display_lr_and_pc`.

lib_standard_app/debug.h
```diff
@@ -1,3 +1,168 @@
...
+
+#ifdef HAVE_LEDGER_ASSERT_DISPLAY
+#ifdef LEDGER_ASSERT_CONFIG_LR_AND_PC_INFO
+#define LR_AND_PC_OFFSET 0
+#define LR_AND_PC_SIZE   30
+WEAK void assert_display_lr_and_pc(int lr, int pc);
+#define ASSERT_DISPLAY_LR_AND_PC(lr, pc) assert_display_lr_and_pc(lr, pc)
+#else
+#define LR_AND_PC_SIZE 0
+#endif
```
xchapron-ledger, line 34: I think we could inverse the `#ifdef` order to make it easier to read.
It would be something like:

```
#if defined(LEDGER_ASSERT_CONFIG_LR_AND_PC_INFO) && defined(HAVE_LEDGER_ASSERT_DISPLAY)
#define LR_AND_PC_OFFSET 0
#define LR_AND_PC_SIZE   30
WEAK void assert_display_lr_and_pc(int lr, int pc);
#define ASSERT_DISPLAY_LR_AND_PC(lr, pc) assert_display_lr_and_pc(lr, pc)
#else
#define LR_AND_PC_SIZE 0
#define ASSERT_DISPLAY_LR_AND_PC(lr, pc) \
    do {                                 \
    } while (0)
#endif

#if defined(LEDGER_ASSERT_CONFIG_LR_AND_PC_INFO) && defined(HAVE_PRINTF)
void print_lr_and_pc(int lr, int pc);
#define ASSERT_PRINT_LR_AND_PC(lr, pc) print_lr_and_pc(lr, pc)
#else
#define ASSERT_PRINT_LR_AND_PC(lr, pc) \
    do {                               \
    } while (0)
#endif
```
how does that sounds to you?

lib_standard_app/debug.c
```diff
@@ -19,20 +19,58 @@
...
-static char errordata[20];
+#if defined(HAVE_LEDGER_ASSERT_DISPLAY) || defined(HAVE_DEBUG_THROWS)
+
+#ifndef ASSERT_BUFFER_LEN
+#define ASSERT_BUFFER_LEN 24
+#endif
+char assert_buffer[ASSERT_BUFFER_LEN];
 
+#ifdef HAVE_DEBUG_THROWS
 WEAK void app_throw_info(unsigned int exception, unsigned int lr_val)
```
xchapron-ledger, line 36: What do you think of dropping all the `WEAK` here? I don't see a reason for an app to change those behaviors. do you?
This might even solve our double `WEAK` issue in `os.c`?

lib_standard_app/debug.c
```diff
@@ -41,15 +79,15 @@ static void review_choice(bool confirm)
...
            {
                .title = "App error",
-               .text  = errordata,
+               .text  = assert_buffer,
            });
 UX_FLOW(ux_error_flow, &ux_error);
 #endif
 
-WEAK void __attribute__((noreturn)) debug_display_throw_error(int exception)
+WEAK void __attribute__((noreturn)) debug_display_exit(int exception)
```
xchapron-ledger, line 90: I think we could drop the parameter no?

xchapron-ledger (commented):
Should allow a potential app callback before displaying the message / exiting the app? something to send back an APDU response in case there was an APDU waiting for a response?
I'd say this could be transparent in standard_app, with a callback in io.c that would respond a standard ASSERT error code + infos?
Or maybe let's keep that for another time. Software wallet probably doesn't need a gentle response when thing goes as bad as an assert is triggered.

lib_standard_app/debug.h
sgliner-ledger: Good idea ! I had a nice formatting but I guess the linter got rid of it :smiling_face_with_tear:

Makefile.standard_app
sgliner-ledger: I was wondering the same thing Letsgo :)

lib_standard_app/debug.h
sgliner-ledger: very good point !

Makefile.standard_app
sgliner-ledger: Well, I am not really fond of this also, but I thought that was what the majority wanted :woman_shrugging: 
I'll just keep it inside debug then :)

src/os.c
```diff
@@ -85,30 +85,22 @@ char os_secure_memcmp(const void *src1, const void *src2, size_t length)
...
-
-#ifdef HAVE_PRINTF
-    PRINTF("exception[%d]: LR=0x%08X\n", exception, lr_val);
-#endif  // HAVE_PRINTF
-
+#ifdef HAVE_DEBUG_THROWS
     // Send to the app the info of exception and LR for debug purpose
-    app_throw_info(exception, lr_val);
+    DEBUG_THROW(exception);
+#endif
```
xchapron-ledger, line 103: Maybe you can keep an `#elif defined HAVE_PRINTF` ?

include/ledger_assert.h
```diff
@@ -0,0 +1,159 @@
+#pragma once
+
+#include "appflags.h"
+#include "bolos_target.h"
+#include "decorators.h"
+#include "errors.h"
+
+void __attribute__((noreturn)) assert_display_exit(void);
+
+#define ASSERT_ERROR_CODE 0xBAAD
```
xchapron-ledger, line 9: Still necessary?

include/ledger_assert.h
```diff
@@ -0,0 +1,159 @@
...
+    } while (0)
+#endif
+
+#ifdef LEDGER_ASSERT_CONFIG_FILE_INFO
+#define LEDGER_ASSERT_FILE_INFO()                     \
+    do {                                              \
+        ASSERT_PRINT_FILE_INFO(__FILE__, __LINE__);   \
+        ASSERT_DISPLAY_FILE_INFO(__FILE__, __LINE__); \
+    } while (0)
+#endif
```
xchapron-ledger, line 120: Couldn't this ones always be defined as the ASSERT_PRINT_XXX and ASSERT_DISPLAY_XXX will be void when not needed?
It would also ease the definition of LEDGER_ASSERT.

src/ledger_assert.c
```diff
@@ -0,0 +1,125 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ ********************************************************************************/
+
+#include "exceptions.h"
+#include "os.h"
+#include <string.h>
```
xchapron-ledger, line 19: At the top as it's a sys/toolchain include?

include/ledger_assert.h
```diff
@@ -0,0 +1,153 @@
+#pragma once
+
+#include "appflags.h"
+#include "bolos_target.h"
+#include "decorators.h"
+#include "errors.h"
+#include "os.h"
```
xchapron-ledger, line 7: Do you really need all of this?

include/ledger_assert.h
```diff
@@ -0,0 +1,153 @@
...
+#define LEDGER_ASSERT_CONFIG_MESSAGE_INFO   1
+#define LEDGER_ASSERT_CONFIG_LR_AND_PC_INFO 1
+#endif
+
+#ifdef LEDGER_ASSERT_CONFIG_MESSAGE_INFO
+#define LEDGER_ASSERT_CONFIG_LR_AND_PC_INFO 1
+#endif
+
+#if defined(LEDGER_ASSERT_CONFIG_LR_AND_PC_INFO) && defined(HAVE_LEDGER_ASSERT_DISPLAY)
+#define LR_AND_PC_OFFSET 0
```
xchapron-ledger, line 19:
```suggestion
```
unused I think

include/ledger_assert.h
```diff
@@ -0,0 +1,153 @@
...
+#define ASSERT_DISPLAY_LR_AND_PC(lr, pc) assert_display_lr_and_pc(lr, pc)
+#else
+#define LR_AND_PC_SIZE 0
+#define ASSERT_DISPLAY_LR_AND_PC(lr, pc) \
+    do {                                 \
+    } while (0)
+#endif
+
+#if defined(LEDGER_ASSERT_CONFIG_MESSAGE_INFO) && defined(HAVE_LEDGER_ASSERT_DISPLAY)
+#define MESSAGE_OFFSET LR_AND_PC_SIZE
```
xchapron-ledger, line 31:
```suggestion
```
unused I think

include/ledger_assert.h
```diff
@@ -0,0 +1,153 @@
...
+#define ASSERT_DISPLAY_MESSAGE(message) assert_display_message(message)
+#else
+#define MESSAGE_SIZE 0
+#define ASSERT_DISPLAY_MESSAGE(message) \
+    do {                                \
+    } while (0)
+#endif
+
+#if defined(LEDGER_ASSERT_CONFIG_FILE_INFO) && defined(HAVE_LEDGER_ASSERT_DISPLAY)
+#define FILE_OFFSET MESSAGE_OFFSET + MESSAGE_SIZE
```
xchapron-ledger, line 43:
```suggestion
```
unused I think

lib_standard_app/main.c
```diff
@@ -72,7 +72,7 @@ static void standalone_app_main(void)
         }
         CATCH_OTHER(e)
         {
-            PRINTF("Exiting following exception: %d\n", e);
+            PRINTF("Exiting following exception: 0x%04X\n", e);
```
xchapron-ledger, line 75: Could you also change the same `printf` in `library_app_main`?

src/ledger_assert.c
```diff
@@ -0,0 +1,144 @@
...
+void throw_display_lr(int e, int lr)
+{
+    lr = compute_address_location(lr);
+    snprintf(assert_buffer, ASSERT_BUFFER_LEN, "e=0x%04X\n LR=0x%08X\n", e, lr);
+}
+
+void throw_print_lr(int lr)
+{
+    lr = compute_address_location(lr);
+    PRINTF("=> LR: 0x%08X \n", lr);
```
xchapron-ledger, line 106: Could be made equivalent to the printf without HAVE_DEBUG_THROWS:
```suggestion
    PRINTF("exception[%d]: LR=0x%08X\n", exception, lr);
```

sgliner-ledger: > Could be made equivalent to the printf without HAVE_DEBUG_THROWS:
The exception code is already printed in the main wether or not you have the define

xchapron-ledger (approved):

PR #488, opened 2023-12-06
Add HAVE_BOLOS_APP_STACK_CANARY by default

fbeutin-ledger:
## Description

Add HAVE_BOLOS_APP_STACK_CANARY by default

## Changes include

- [ ] Tests
- [ ] Documentation


Makefile.defines
```diff
@@ -52,6 +52,11 @@ endif
 # APPNAME exposed to the app as a CFLAG because it might contain spaces
 CFLAGS += -DAPPNAME=\"$(APPNAME)\"
 
+# Stack canary enabled by default
+ifndef NO_BOLOS_APP_STACK_CANARY
```
xchapron-ledger, line 56:
```suggestion
ifndef DISABLE_BOLOS_APP_STACK_CANARY
```

xchapron-ledger:
This looks good, I'm wondering if we should go a big further with:
- Exposing a `APP_STACK_CANARY_CHECK` macro?
- Considering changing the `io_seproxyhal_se_reset();` into a `LEDGER_ASSERT`? This would help debugging, but is it a good idea if the stack have overflow to continue doing thing that are done when a `LEDGER_ASSERT` is called?

PR #500, opened 2024-01-10
src: Temporary patch for cx_crc32

xchapron-ledger:
## Description

Current CX lib CRC32 implementation is buggy.
So fix it with software implementation to be embedded in app binary until the Os version is fixed.

## Changes include

- [ ] Tests
- [ ] Documentation


lpascal-ledger (approved):
LGTM, but may benefit from a review from @srasoamiaramanana-ledger

srasoamiaramanana-ledger (approved):

PR #501, opened 2024-01-11
Support new font sizes

nroggeman-ledger:
## Description

The goal of this PR is to support new font sizes (28 & 36) based on Inter Regular and Inter Semibold fonts, for later usage. 
Because these sizes can be bigger, we have to change the format of internal font structures, so also the tool to generate fonts.
And instead of using directly the font names (containing the size), we will now use font alias (small, large...).

## Changes include

- [ ] Tests
- [ ] Documentation


xchapron-ledger (commented):
Small question not really regarding this PR, but would it be possible to move the nbgl_font_*.inc file in a dedicated folder? There are many of them and it doesn't help having a great view of the include folder with include file usable by apps.

abonnaudet-ledger (approved):

PR #504, opened 2024-01-15
LEDGER_ASSERT improvments and CX_ASSERT addition

xchapron-ledger:
## Description

Add some documentation for `LEDGER_ASSERT`
Allow `printf` like usage of `LEDGER_ASSERT` for error message
Minor cleanup on CX includes
Add `CX_ASSERT` macro

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/include/cx_errors.h
+++ b/include/cx_errors.h
@@ -6,6 +6,7 @@
 #pragma once
 
 #include <stdint.h>
+#include "ledger_assert.h"
 
 /**
  * Checks the error code of a function.
@@ -30,6 +31,15 @@
         if (error && error != CX_CARRY) { \
             goto end;                     \
         }                                 \
+    } while (0)
+
+/**
+ * Checks the error code of a function and assert in case of error.
+ */
+#define CX_ASSERT(call)                                                   \
+    do {                                                                  \
+        cx_err_t _assert_err = call;                                      \
+        LEDGER_ASSERT(!_assert_err, "err 0x%X <%s>", _assert_err, #call); \
     } while (0)
 
 /** Success. */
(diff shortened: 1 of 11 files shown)


xchapron-ledger:
For future, here is the script used to generate most of the PR (except includes and formatting):
```
for file in files:
    print("Working on:", file)

    with open(file, "r") as f:
        iterator = iter(f.readlines())

    with open(file, "w") as fout:
        line = next(iterator)
        while line:
            content = None

            if line.startswith("WARN_UNUSED_RESULT cx_err_t"):
                content = line
                while ";" not in content:
                    content += next(iterator)

            if content:
                fout.write(content)
                fout.write("\n")

                starters, params = content.split("(")

                # Extract params names and create a string with the list without the types
                assert params.endswith(");\n")
                params = params.replace(");\n", "")
                params_list = params.replace("\n", " ").replace("*", "").split(",")
                params_value = ", ".join([x.strip().split(" ")[-1] for x in params_list])

                # Extract func_name and create new_func_name
                func_name = starters.replace("\n", " ").split(" ")[-1]
                if func_name.endswith("_no_throw"):
                    new_func_name = func_name.replace("_no_throw", "_assert")
                else:
                    new_func_name = func_name + "_assert"

                fout.write(f"/**\n * @brief {func_name} version which doesn't return in case of failure.\n *\n * See #{func_name}\n */")
                fout.write("\n")

                fout.write("static inline void " + new_func_name + "(")
                fout.write(params)
                fout.write(")\n")

                fout.write("{\n")
                fout.write(f"    LEDGER_ASSERT({func_name}({params_value}) == CX_OK,\n")
                fout.write(f"                  \"{func_name}\");\n")
                fout.write("}\n")
            else:
                fout.write(line)

            try:
                line = next(iterator)
            except StopIteration:
                print("End of file")
                line = None
```

PR #505, opened 2024-01-16
ledger_assert: add header for HAVE_PRINTF

sgliner-ledger:
## Description

Add `os.h` include when using `HAVE_PRINTF` for ledger_assert

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/include/ledger_assert.h
+++ b/include/ledger_assert.h
@@ -2,6 +2,10 @@
 
 #include <stdbool.h>
 
+#ifdef HAVE_PRINTF
+#include "os.h"
+#endif
+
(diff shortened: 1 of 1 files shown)


include/ledger_assert.h
```diff
@@ -2,6 +2,10 @@
 
 #include <stdbool.h>
 
+#ifdef HAVE_PRINTF
+#include <os.h>
```
xchapron-ledger, line 6:
```suggestion
#include "os.h"
```
Cause it's not a toolchain header but a SDK one?
Though we don't follow any strict rules about this.

cedelavergne-ledger (approved):

xchapron-ledger (approved):

PR #549, opened 2024-02-28
Add the europa target in the build CI

tdejoigny-ledger:


xchapron-ledger (commented):
Would be nice to rework this and to use something like:
- list all repositories in LedgerHQ
- keep only the repos with naming like app-*
- keep only the apps that are not archived
- keep only the apps with a `ledger_app.toml` file
- keep only the apps with `sdk = "C"` in the `ledger_app.toml` file

And then:
- use the `ledger_app.toml` file `build_directory` param instead of hardcoded `${{ env.build_path }}`
- use the `ledger_app.toml` file `devices` param instead of the hardcoded `${{ env.build_for_stax == 'true' }}`

xchapron-ledger:
europa could be added here too: https://github.com/LedgerHQ/ledger-secure-sdk/blob/master/.github/workflows/check_clang_static_analyzer.yml#L16

cedelavergne-ledger (dismissed):

tdejoigny-ledger:
> europa could be added here too: https://github.com/LedgerHQ/ledger-secure-sdk/blob/master/.github/workflows/check_clang_static_analyzer.yml#L16

OK

tdejoigny-ledger:
> Would be nice to rework this and to use something like:
> 
> * list all repositories in LedgerHQ
> * keep only the repos with naming like app-*
> * keep only the apps that are not archived
> * keep only the apps with a `ledger_app.toml` file
> * keep only the apps with `sdk = "C"` in the `ledger_app.toml` file
> 
> And then:
> 
> * use the `ledger_app.toml` file `build_directory` param instead of hardcoded `${{ env.build_path }}`
> * use the `ledger_app.toml` file `devices` param instead of the hardcoded `${{ env.build_for_stax == 'true' }}`

@xchapron-ledger  here you mean that we should build the list dynamically ?

xchapron-ledger:
> @xchapron-ledger here you mean that we should build the list dynamically ?

Yep, So that it's always up-to-date!

tdejoigny-ledger:
> > @xchapron-ledger here you mean that we should build the list dynamically ?
> 
> Yep, So that it's always up-to-date!

@xchapron-ledger I reworked the CI as per your suggestion. I have implemented it in another [PR](https://github.com/LedgerHQ/ledger-secure-sdk/pull/553)

Still many apps don't have the manifest file....

xchapron-ledger:
> Still many apps don't have the manifest file....

I'd say we can either consider that apps without manifest are C apps with only LNS / LNX / LNSP support, or just don't consider them for the CI while making sure apps available on P1 have the manifest? I think there are not that much missing?

cedelavergne-ledger (approved):

xchapron-ledger:
Already done in 2e2fec207e93c918350f665a96768f5cfa7b3b11

PR #553, opened 2024-03-04
Update the apps build CI (use apps manifests)

tdejoigny-ledger:
## Description

Update the apps build CI to get all the apps repos and to use the manifest files to build each app

### DIFF
--- a/.github/workflows/build_all_apps.yml
+++ b/.github/workflows/build_all_apps.yml
@@ -32,7 +32,7 @@ jobs:
           page=1
           result=()
           while true; do
-            repos=$(curl -s "https://api.github.com/orgs/LedgerHQ/repos?page=$page&per_page=200")
+            repos=$(curl -s "https://api.github.com/orgs/LedgerHQ/repos?page=$page&per_page=200&type=public")
 
             if ! echo "$repos" | jq -e '.[].name' >/dev/null; then
               # all repos have been listed
@@ -41,7 +41,7 @@ jobs:
             fi
 
             # add filtered repos to the list
-            for repo in $(echo "$repos" | jq -r '.[] | select((.name | startswith("app-")) and (.archived == false) and (.private == false)) | .name'); do
+            for repo in $(echo "$repos" | jq -r '.[] | select((.name | startswith("app-")) and (.archived == false)) | .name'); do
               default_branch=$(echo "$repos" | jq -r --arg repo "$repo" '.[] | select(.name == $repo) | .default_branch')
               if [[ ! " ${BLACKLIST//,/ } " =~ " $repo " ]]; then
                 result+=("$repo,$default_branch")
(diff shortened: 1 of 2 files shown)


.github/workflows/build_all_apps.yml
```diff
@@ -3,313 +3,91 @@ name: Build all apps on latest SDK
...
+          done
+
+          echo "repo_list=${result[@]}" >> $GITHUB_ENV
+
+      - name: Extract metadata
+        id: set-matrix
+        run: |
+          repo_info="["
+          for repo in ${{ env.repo_list }}; do
+            toml_content=$(curl -s "https://raw.githubusercontent.com/LedgerHQ/$repo/develop/ledger_app.toml")
```
xchapron-ledger, line 54: Is there a way to not explicitly use `develop` here?
Or this should at least be also enforced in `Clone App` step below as `develop` might not be the default branch.

tdejoigny-ledger: @xchapron-ledger I slightly modified the CI workflow by first retrieving the default branch, then using this branch to fetch the manifest, and finally cloning and building from it.

xchapron-ledger (commented):
I think we should use more ledgered instead of manually parsing the manifest here.

tdejoigny-ledger:
> I think we should use more ledgered instead of manually parsing the manifest here.

We briefly discussed with @lpascal-ledger , for the manifest parsing part, it's not necessarily clear that ledgered is relevant here. Additionally, ledgered takes a file path as an argument, but here I retrieve the file's content.

I'm wondering if it's not relevant to first create this version before having a more advanced version in Ledgered that combines repository search and manifest parsing.

lpascal-ledger:
While I agree this type of PR would be better of with `ledgered`, as of today it would not shorten this particular one.
So either we wait for a week or two (or more) for `ledgered` to be up to the task, or we merge early.
:shrug:

.github/workflows/build_all_apps.yml
```diff
@@ -16,300 +11,94 @@ on:
...
+      - name: Extract metadata
+        id: set-matrix
+        run: |
+          repo_info="["
+          for repo in ${{ env.repo_list }}; do
+            repo_name=$(echo "$repo" | cut -d ',' -f 1)
+            default_branch=$(echo "$repo" | cut -d ',' -f 2)
+            toml_content=$(curl -s "https://raw.githubusercontent.com/LedgerHQ/$repo_name/$default_branch/ledger_app.toml")
+            # check if the toml file is present in the repo and if the repo is not for a rust app
+            if [[ ! $toml_content =~ "404: Not Found" ]] && [[ ! $toml_content =~ "rust-app" ]]; then
```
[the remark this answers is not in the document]
tdejoigny-ledger: ok done

lpascal-ledger (approved):

PR #554, opened 2024-03-04
Enable CRC hardware implementation

srasoamiaramanana-ledger:
## Description

This PR implements a new syscall for CRC computations.

## Changes include

- [ ] Tests
- [ ] Documentation

## Breaking changes

The syscall cx_crc32_hw has been removed and replaced by cx_crc_hw which allows for both CRC16 and CRC32 computations.


src/cx_stubs.S
```diff
@@ -56,9 +56,7 @@ CX_TRAMPOLINE _NR_cx_cmac                                  cx_cmac
 CX_TRAMPOLINE _NR_cx_constant_time_eq                      cx_constant_time_eq
 CX_TRAMPOLINE _NR_cx_crc16                                 cx_crc16
 CX_TRAMPOLINE _NR_cx_crc16_update                          cx_crc16_update
-// Disable lib CX CRC32 until it get fixed in the OS.
-// Meanwhile, it's implemented in cx_wrappers.c
-//CX_TRAMPOLINE _NR_cx_crc32                                 cx_crc32
+CX_TRAMPOLINE _NR_cx_crc32                                 cx_crc32
```
xchapron-ledger, line 59: Is it worth it to have simple `syscall` wrapper like this in the `cx_lib`? Could be simple `static inline` wrapper in the `.h`?
Also should be expose a `cx_crc32_update` too?

yhql: I guess we can now replace `cx_crc16` and `cx_crc32` with a single exposed syscall to `cx_crc`, and have wrappers for 16 or 32 versions

yhql (approved):

PR #563, opened 2024-03-08
Split NBGL layout layer

nroggeman-ledger:
## Description

The goal of this PR is to split  NBGL layout layer implementation, because it's getting to big to fit a single file.
The implementation for Nano and Wallet have to be split, and then the implementations of Keyboard, Keypad and Navigation bar.

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/lib_nbgl/include/nbgl_layout.h
+++ b/lib_nbgl/include/nbgl_layout.h
@@ -26,8 +26,7 @@ extern "C" {
 /*********************
  *      DEFINES
  *********************/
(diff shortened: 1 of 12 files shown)


xchapron-ledger (approved):
Tested OK on app-boilerplate.
Only visual change vs master is the QRcode screen which gets fixed with this version.

abonnaudet-ledger (approved):

abonnaudet-ledger:
Tested :ok: on OS

PR #573, opened 2024-03-18
Generate glyps for apps screenshots

nroggeman-ledger:
## Description

The goal of this PR is to be able to generate screenshots from bolos-ng, but with fake apps, to compare apps screenshots with Figma ones.
To do so, it's necessary to generate an apps_glyphs.c/apps_glyphs.h pair.
Moreover, some minor improvements have been done in NBGL to improve similarities with Figma

## Changes include

- [ ] Tests
- [ ] Documentation


lib_nbgl/src/nbgl_use_case.c
```diff
@@ -1599,7 +1600,9 @@ void nbgl_useCaseGenericSettings(const char                   *appName,
     pageTitle = appName;
     navType   = GENERIC_NAV;
 
-    memcpy(&genericContext.genericContents, settingContents, sizeof(nbgl_genericContents_t));
+    if (settingContents != NULL) {
+        memcpy(&genericContext.genericContents, settingContents, sizeof(nbgl_genericContents_t));
+    }
```
xchapron-ledger, line 1605: Humf I have indeed not handled this case!
I think I will do it in another PR cause the proper fix is probably different?

nroggeman-ledger: I've not made 100s of tests but with this fix the top-right icon is OK, and the infos are properly displayed... Maybe there is a simpler one, but it seems difficult...

xchapron-ledger: Ok, I was afraid that the info display content won't work, but it does indeed, so this is ok for me!

xchapron-ledger (approved):

xchapron-ledger:
Tested ok sur app-boilerplate.

lib_nbgl/src/nbgl_use_case.c
```diff
@@ -2431,7 +2434,7 @@ void nbgl_useCaseAddressReview(const char                      *address,
 #ifdef TARGET_STAX
     centeredInfo->text3 = NULL;
 #else   // TARGET_STAX
-    centeredInfo->text3               = "Swipe to review";
+    centeredInfo->text3               = "Swipe to continue";
```
xchapron-ledger, line 2437: Is this wording specific to address review?
Else "Swipe to review" must be change in others places in this file.

PR #579, opened 2024-03-21
Add osu defines when needed

tjoly-ledger:
## Description

Add OSU definitions when needed

## Changes include

- [X ] New feature (non-breaking change that adds functionality)

### DIFF
--- a/include/os_io.h
+++ b/include/os_io.h
@@ -1,6 +1,7 @@
 #pragma once
 #include <stdbool.h>
 #include <stdint.h>
+#include "app_config.h"
 #include "bolos_target.h"
 #include "os_math.h"
 #include "decorators.h"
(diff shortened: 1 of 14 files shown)


xchapron-ledger:
IMO this solution seem fragile, when updating SDK files, we are going to often break the OSU build as OSU build is not check by the SDK CI...

tjoly-ledger:
> IMO this solution seem fragile, when updating SDK files, we are going to often break the OSU build as OSU build is not check by the SDK CI...

Use existing flag BOLOS_OS_UPGRADER_APP instead of OSU to not introduce an other FLAG.

jarevalo-ledger (approved):
LGTM

include/os_io.h
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#include "app_config.h"
```
xchapron-ledger, line 2: Could be below toolchain includes and just above `#include "bolos_target.h"`?

tjoly-ledger: done

xchapron-ledger (approved):

PR #588, opened 2024-03-26
Embed graphic library name into binary

lpascal-ledger:


xchapron-ledger (approved):
Did you try reading it from the elf to check it's indeed correctly added and encoded?

PR #630, opened 2024-04-22
Release API LEVEL 15 v15.4.0

fbeutin-ledger:


xchapron-ledger (approved):
Should be checked with @abonnaudet-ledger and @nroggeman-ledger if all NBGL fixes included should indeed be included?

cedelavergne-ledger (approved):

PR #634, opened 2024-04-25
Keep pointer for nbgl_layoutAddKeypad (LNX/LNS+)

dmorais-ledger:
## Description

This PR let the LNX/LNS+ version of **nbgl_layoutAddKeypad** use a pointer like before, and a string ID just when generating screenshots.

## Changes include


lib_nbgl/include/nbgl_layout.h
```diff
@@ -666,11 +666,15 @@ int            nbgl_layoutUpdateKeypadContent(nbgl_layout_t *layout,
...
                                               const char    *text);
 
-#else   // HAVE_SE_TOUCH
+#else  // HAVE_SE_TOUCH
 /* layout objects for pages with keypad (nanos) */
 int nbgl_layoutAddKeypad(nbgl_layout_t       *layout,
                          keyboardCallback_t   callback,
+#ifdef BUILD_SCREENSHOTS
                          UX_LOC_STRINGS_INDEX textId,
+#else   // BUILD_SCREENSHOTS
```
nroggeman-ledger, line 675: I'm not OK with this, the ID shall be added, not replacing text....

nroggeman-ledger (changes requested):

lib_nbgl/src/nbgl_fonts.c
```diff
@@ -772,10 +772,12 @@ uint16_t nbgl_getTextNbLinesInWidth(nbgl_font_id_e fontId,
                 if (last_nb_lines < nbLines) {
                     last_nb_lines = nbLines;
                 }
+#ifndef SCREEN_SIZE_WALLET
```
nroggeman-ledger, line 775: Better use #ifdef SCREEN_SIZE_NANO than #ifndef SCREEN_SIZE_WALLET

Unfortunately it's not enough, the "if (with != 0)" must also be surrounded with #ifdef SCREEN_SIZE_NANO... but why is the interrest of this condition ?

nroggeman-ledger (changes requested):

lib_nbgl/src/nbgl_fonts.c
dmorais-ledger: The change related to **_if (nbLines == 4)_** is only intended to nanos devices when generating screenshots: thanks for the **#ifdef SCREEN_SIZE_NANO**, I will use it.
The **_if (with != 0)_** is intended to fix an issue about how nb lines is computed: I can make it nano devices specific, if you want, but I think we need to change the way the number of lines is computed and this is the first step in that direction.

nroggeman-ledger: let's discuss it later but yes, make it Nano Specific because it creates a bug in NOMINAL Stax/Flex UX

dmorais-ledger: Ok, I reverted the changes I made with **if (with != 0)** unless where it is necessary for LNX/LNS+ and we'll discuss about that later.

nroggeman-ledger (approved):

PR #639, opened 2024-04-29
Improve CX_CHECK

cedelavergne-ledger:
Add `PRINTF` inside `CX_CHECK` to ease debug, and find the file/line of the error

### DIFF
--- a/include/cx_errors.h
+++ b/include/cx_errors.h
@@ -12,12 +12,13 @@
  * Checks the error code of a function.
  * @hideinitializer
  */
-#define CX_CHECK(call) \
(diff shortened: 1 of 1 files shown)


xchapron-ledger (approved):

xchapron-ledger (commented):
This might cost a lot of flash while in debug, but I guess we don't care much...

cedelavergne-ledger:
> This might cost a lot of flash while in debug, but I guess we don't care much...

Yes, but it is only when `DEBUG` is active. Should not be the case by default

PR #679, opened 2024-06-03
Application storage

jarevalo-ledger:
## Description

Application storage rework.

## Changes include

- [X ] New feature (non-breaking change that adds functionality)
- [ ] Tests
- [ ] Documentation


include/app_storage.h
```diff
@@ -0,0 +1,81 @@
...
+#define APP_STORAGE_PROP_SETTINGS (1 << 0)
+///< bit to indicate in properties that the application contains data
+#define APP_STORAGE_PROP_DATA     (1 << 1)
+
+/**
+ * @brief Structure defining the header of NVRAM
+ *
+ */
+typedef struct app_storage_header_s {
+    char     tag[4];          ///< ['S','T','O','R'] array, when properly initialized
```
cedelavergne-ledger, line 37: You have used NVRA in the code below...

include/app_storage.h
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * @brief Structure defining the NVRAM
+ *
+ */
+typedef struct app_storage_s {
+    app_storage_header_t header;  ///< header describing the data
+#ifndef HAVE_BOLOS
+    app_storage_data_t
+        data;  ///< application data, app_storage_data_t must be defined in "nvram_data.h" file
```
cedelavergne-ledger, line 55: Why defining it in nvram_data.h? It could be a simple pointer, no?
What happens if the app doesn't define any structure then?

include/app_storage.h
jarevalo-ledger: Fixed

include/app_storage.h
jarevalo-ledger: Fixed, it is `app_storage_data.h`.
The struct `app_storage_t` is actually written in NVRAM, thus we need to have it contiguous in memory (header+data).
If an application enable in its Makefile the app storage (`ENABLE_APP_STORAGE`), then it must provide a header file named `app_storage_data.h` that defines this `app_storage_data_t` struct otherwise it will not compile.

cedelavergne-ledger: Thanks.
Is this already defined somewhere in a doc? At least on Confluence, maybe also in Dev Portal?

cedelavergne-ledger (approved):

abonnaudet-ledger (approved):

fbeutin-ledger (approved):

Makefile.standard_app
```diff
@@ -47,6 +47,14 @@ ifeq ($(ENABLE_SWAP), 1)
     DEFINES += HAVE_SWAP
 endif
 
+#####################################################################
+#                              APP STORAGE                          #
+#####################################################################
+ifeq ($(ENABLE_APP_STORAGE), 1)
```
fbeutin-ledger, line 53: Small nitpick is that this option is not very self explanatory
Maybe something like `ENABLE_OS_MANAGED_APP_STORAGE`, or `ENABLE_UPDATE_RESILIENT_APP_STORAGE` or `ENABLE_PERSISTENT_APP_STORAGE`

Makefile.standard_app
```diff
@@ -47,6 +47,14 @@ ifeq ($(ENABLE_SWAP), 1)
...
+    HAVE_APP_STORAGE = 1
```
fbeutin-ledger, line 54: Same as above

PR #687, opened 2024-06-10
Clarify NFC default setting

abonnaudet-ledger:
## Description

- `HAVE_NFC_ACTIVATED_BY_DEFAULT` used to change the interpretation of _NFC enabled_ setting. This PR clarifies the behavior of the setting value.
- Add some helper defines 

## Changes include

- [ ] Tests
- [ ] Documentation


include/os_settings.h
```diff
@@ -112,13 +112,26 @@ typedef enum os_setting_e {
  *
  */
 // if (os_settings[OS_SETTING_FEATURES] & OS_SETTING_FEATURES_NFC_ENABLED) then NFC is enabled
-#define OS_SETTING_FEATURES_NFC_ENABLED    0x1
+#define OS_SETTING_FEATURES_NFC_ENABLED_MASK    0x0
```
nroggeman-ledger, line 115: It's not a mask but an offset, the mask is actually (1 << 0), so it should be renammed to OS_SETTING_FEATURES_NFC_ENABLED_OFFSET

nroggeman-ledger (changes requested):

nroggeman-ledger (approved):

PR #727, opened 2024-07-16
Add up footer layout object and tip-box object

nroggeman-ledger:
## Description

The original goal of this PR is to support Tip-Box, as on the attached picture (from https://www.figma.com/design/k9Q6hP0XWAw00AiT6GDb0O/02.-App-SDK?node-id=94-69834&t=ZUjo31oi2l0Y5sc1-4)
image

To do so, it's necessary to add a up-footer object (and container) in layout layer. This addition enable to regroup other "similar" objects in this up-footer object, like long-press button, horizontal buttons and simple button (in case of button in bottom)

Moreover, to link the content of the tip-box to the centered info), an extended centered info has to be created.

Finally, an nbgl_useCaseAdvancedReview() API is created to be able to use this tip-box in the case illustrated by the attached picture

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/lib_nbgl/include/nbgl_use_case.h
+++ b/lib_nbgl/include/nbgl_use_case.h
@@ -151,6 +151,21 @@ typedef struct {
 } nbgl_homeAction_t;
 
 /**
+ * @brief The necessary parameters to build a tip-box in first review page and
+ * the modal if this tip box is touched
+ *
+ */
+typedef struct {
+    const char                *text;  ///< text of the tip-box
+    const nbgl_icon_details_t *icon;  ///< icon of the tip-box
+    const char *modalTitle;   ///< title given to modal window displayed when tip-box is touched
+    nbgl_contentType_t type;  ///< type of page content in the following union
+    union {
+        const nbgl_contentInfoList_t infos;  ///< infos pairs displayed in modal.
+    };
+} nbgl_tipBox_t;
+
+/**
  * @brief The different types of operation to review
  *
  */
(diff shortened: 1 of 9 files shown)


lib_nbgl/include/nbgl_use_case.h
```diff
@@ -150,6 +150,18 @@ typedef struct {
...
+/**
+ * @brief The necessary parameters to build a tip-box in first review page and
+ * the modal if this tip box is touched
+ *
+ */
+typedef struct {
+    const char                *text;  ///< text of the tip-box
+    const nbgl_icon_details_t *icon;  ///< icon of the tip-box
+    const char *modalTitle;  ///< title given to modal window displayed when tip-box is touched
+    const nbgl_contentInfoList_t infos;  ///< infos pairs displayed in modal.
```
cedelavergne-ledger, line 162: Could we already extend this, for example to allow displaying something like _Center Text_, to just give more informatio, display address, accound name.... ?
It could be just using a union, to be completed when needed?


typedef struct {
    const char                *text;  ///< text of the tip-box
    const nbgl_icon_details_t *icon;  ///< icon of the tip-box
    const char *modalTitle;  ///< title given to modal window displayed when tip-box is touched
    union {
        const nbgl_contentInfoList_t infos;  ///< infos pairs displayed in modal.
    } content;
} nbgl_tipBox_t;

lib_nbgl/src/nbgl_layout.c
```diff
@@ -3024,6 +2903,275 @@ int nbgl_layoutAddExtendedFooter(nbgl_layout_t *layout, const nbgl_layoutFooter_
...
+    }
+    if ((upFooterDesc == NULL) || (upFooterDesc->type >= NB_UP_FOOTER_TYPES)) {
+        return -2;
+    }
+
+    layoutInt->upFooterContainer
+        = (nbgl_container_t *) nbgl_objPoolGet(CONTAINER, layoutInt->layer);
+    layoutInt->upFooterContainer->obj.area.width = SCREEN_WIDTH;
+    layoutInt->upFooterContainer->layout         = VERTICAL;
+    // maximum 4 children for long press button
```
cedelavergne-ledger, line 2934: 4 for long press. Is it the absolute maximum, whatever the types?
What about other type max children?

lib_nbgl/include/nbgl_use_case.h
nroggeman-ledger: Indeed, I will rework it that way, with an additionnal field to indicate the type used in the union

lib_nbgl/src/nbgl_layout.c
nroggeman-ledger: 4 is the worst case in the switch below. Then we allocate an array of 4 children, before the switch, and we are sure that it will be suffisiant

nroggeman-ledger: There is already a comment :` // maximum 4 children for long press button`

cedelavergne-ledger (approved):

PR #728, opened 2024-07-16
Fido: Remove old usb hack

yrichard-ledger:
Don't re-configure u2f endpoints every time the host sends a GetDescriptor

## Description

FIDO over USB crashes on android because Android makes GetDescriptor requests between authenticator operations. 
This PR fixes that issue by removing the old hacky code that was present in the `GetReportDescriptor_impl` function.

## Changes include

- [ ] Tests
- [ ] Documentation


lpascal-ledger (approved):
I can't say if this is technically correct, but it seems to correctly fix the faulty behavior.

PR #759, opened 2024-09-03
Fix replacement of last char by "..." when using nbgl_textWrapOnNbLines()

nroggeman-ledger:
## Description

The goal of this PR is to fix an issue in nbgl_textWrapOnNbLines().
This function is intended to create a multi-lines text fitting in the given width and number of lines, wrapping on spaces, and, if the text is too long, the last char is replaced by "...". 
But sometimes, the "..." is too long and the last line doesn't fit in width, so in these case the last 2 characters must be replaced by "..."

## Changes include

- [ ] Tests
- [ ] Documentation


ndedouhet-ledger (approved):
LGTM

tdejoigny-ledger (approved):

PR #768, opened 2024-09-17
Use nbgl shared functions

nroggeman-ledger:
## Description

The goal of this PR is to reduce the size in Flash (and RAM) of Applications, by sharing all low-level functions of NBGL with the OS:
- Object management
- Touchscreen management
- Screens management 

The architecture/Design document can be found in https://ledgerhq.atlassian.net/wiki/spaces/FW/pages/4750966807/BOLOS+Apps+-+Arch+-+NBGL+sharing+between+FW+and+Apps

## Changes include

- [ ] Tests
- [ ] Documentation

## Breaking changes

For apps non using standard makefiles, there may be some risks if NBGL low-level files are compiled.
Speculos won't work. Use API LEVEL 22


lib_nbgl/src/nbgl_obj.c
```diff
@@ -1639,12 +1640,17 @@ void nbgl_refreshReset(void)
 
 /**
  * @brief This functions inits all internal of nbgl objects layer
+ * @note it is supposed to be called only from App
```
cedelavergne-ledger, line 1643: Does it means we'll have to modify all the different Apps to add such call?

lib_nbgl/src/nbgl_touch.c
```diff
@@ -255,29 +265,41 @@ static nbgl_touchType_t nbgl_detectSwipe(nbgl_touchStatePosition_t *last,
 /**********************
  *   GLOBAL FUNCTIONS
  **********************/
+/**
+ * @brief Function to initialize the touch context
+ * @param fromUx if true, means to initialize the UX context, otherwise App one
+ */
+void nbgl_touchInit(bool fromUx)
```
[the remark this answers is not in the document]
nroggeman-ledger: For NBGL touch/screen management, there can be only two possibilities. We are in OS/UX context or not... so a boolean makes sense

lib_nbgl/src/nbgl_obj.c
nroggeman-ledger: This function is already called by all apps, through UX_INIT()

lib_nbgl/include/nbgl_touch.h
```diff
@@ -23,18 +23,19 @@ extern "C" {
...
  **********************/
-void        nbgl_touchHandler(nbgl_touchStatePosition_t *touchEvent, uint32_t currentTimeMs);
-bool        nbgl_touchGetTouchedPosition(nbgl_obj_t                 *obj,
-                                         nbgl_touchStatePosition_t **firstPos,
-                                         nbgl_touchStatePosition_t **lastPos);
+void nbgl_touchInit(bool fromUx);
+void nbgl_touchHandler(bool fromUx, nbgl_touchStatePosition_t *touchEvent, uint32_t currentTimeMs);
+bool nbgl_touchGetTouchedPosition(nbgl_obj_t                 *obj,
+                                  nbgl_touchStatePosition_t **firstPos,
+                                  nbgl_touchStatePosition_t **lastPos);
```
yrichard-ledger, line 38: Why do we need to have separate contexts for ux and apps ?

nroggeman-ledger: As explained in the architecture document, UX and App are now sharing many variables, but for the communication between App and UX (event forwarding and ability to use events in App), it's sometime necessary to know the number of screens consumed by UX, or, for Touch management, to have separated contexts, to compute NBGL Touch Events (touch duration for example)

abonnaudet-ledger (approved):

cedelavergne-ledger (approved):

PR #793, opened 2024-10-07
Full clang build for standard apps

apaillier-ledger:
## Description

Now only uses clang for all steps of the build instead of the assembly of .S files and binary linking being delegated to GCC.

## Changes include

- [ ] Tests
- [ ] Documentation


apaillier-ledger:
One point of attention is the branch instruction `b` from .S files is assembled differently
* with GCC : `b.n` 2 bytes
* with clang : `b.w` 4 bytes

PR #797, opened 2024-10-14
Create a QR Code type of modal window for Tag/Value alias

nroggeman-ledger:
## Description

The goal of this PR is to create a QR Code type of modal window for Tag/Value alias

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/lib_nbgl/src/nbgl_use_case.c
+++ b/lib_nbgl/src/nbgl_use_case.c
@@ -1146,12 +1146,24 @@ static void displayFullValuePage(const nbgl_contentTagValue_t *pair)
                                                   .backAndText.token  = 0,
                                                   .backAndText.tuneId = TUNE_TAP_CASUAL,
                                                   .backAndText.text   = PIC(pair->item)};
-    const char              *info;
-    genericContext.modalLayout = nbgl_layoutGet(&layoutDescription);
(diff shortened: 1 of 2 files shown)


lib_nbgl/src/nbgl_use_case.c
```diff
@@ -1146,28 +1146,36 @@ static void displayFullValuePage(const nbgl_contentTagValue_t *pair)
...
-    const char              *info;
-    genericContext.modalLayout = nbgl_layoutGet(&layoutDescription);
+    genericContext.modalLayout                 = nbgl_layoutGet(&layoutDescription);
     // add header with the tag part of the pair, to go back
     nbgl_layoutAddHeader(genericContext.modalLayout, &headerDesc);
-    // add full value text
-    if (pair->extension->explanation == NULL) {
+    // add either QR Code or full value text
+    if (pair->extension->aliasType == QR_CODE_ALIAS) {
+#ifdef NBGL_QRCODE
```
tdejoigny-ledger, line 1154: should we move the #ifdef before the if and #endif between the else and "{" ?

nroggeman-ledger: Honnestly I prefer not to, because of the else condition... if used without flag, il will simply display nothing, which is better

cedelavergne-ledger (approved):

PR #822, opened 2024-11-13
add nbgl tests workflow

tdejoigny-ledger:


.github/workflows/nbgl_tests.yml
```diff
@@ -0,0 +1,24 @@
...
+    branches:
+      - master
+      - main
+      - develop
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  run_NBGL_tests:
+    runs-on: ubuntu-latest
```
cedelavergne-ledger, line 14: Would it work more simply using something like:

```
jobs:
  run_NBGL_tests:
    uses: LedgerHQ/app-nbgl-tests/.github/workflows/build_and_functional_tests.yml@master
    with:
      golden_run: "Open a PR"
    secrets: inherit
```
:thinking:

tdejoigny-ledger: yes could be interesting if it's working and if the PR can be opened in the remote repository

tdejoigny-ledger: https://github.com/LedgerHQ/ledger-secure-sdk/actions/runs/12199393685
It seems that this option is not working

PR #827, opened 2024-11-25
NBGL extensions for future UX screens

nroggeman-ledger:
## Description

The goal of this PR is to add some features to NBGL to enable creating (or extending) future Bolos UX screens.
These extensions are:
- Support animation in centered info (instead icon)
- Support dual rouded buttons in footer
- Support an additionnal icon in some header
- Support sub-text in spinner
- Support ability to change validation icon type in PinPad with arrow instead of check

## Changes include

- [ ] Tests
- [ ] Documentation


lib_nbgl/src/nbgl_layout.c
```diff
@@ -726,9 +792,9 @@ static nbgl_container_t *addContentCenter(nbgl_layoutInternal_t      *layoutInt,
 
     // get container children
     container->nbChildren = 0;
-    container->children   = nbgl_containerPoolGet(5, layoutInt->layer);
+    container->children   = nbgl_containerPoolGet(6, layoutInt->layer);
```
cedelavergne-ledger, line 795: Could be interesting to have defines instead of hardcoded values. Here, it is difficult to understand

lib_nbgl/src/nbgl_layout.c
```diff
@@ -448,13 +448,79 @@ static void spinnerTickerCallback(void)
...
+    }
+
+    // get index of image obj
+    while (i < layout->container->nbChildren) {
+        if (layout->container->children[i]->type == CONTAINER) {
+            nbgl_container_t *container = (nbgl_container_t *) layout->container->children[i];
+            if (container->children[1]->type == IMAGE) {
+                image = (nbgl_image_t *) container->children[1];
+                if (layout->animation->parsing == LOOP_PARSING) {
+                    if (layout->iconIdxInAnim == (layout->animation->nbIcons - 1)) {
```
cedelavergne-ledger, line 492: This could be a simple ++, with a modulo...

lib_nbgl/src/nbgl_layout.c
nroggeman-ledger: I'll add some details about why 6 :)

lib_nbgl/src/nbgl_layout.c
nroggeman-ledger: Not sure that it would be more readable nor efficient (because not a ^2 static modulo)

abonnaudet-ledger (approved):

PR #844, opened 2025-01-23
Using reusable workflows to trigger all C apps build

lpascal-ledger:


tdejoigny-ledger (commented):
Good simplification! 
Is it possible to display more clearly in the CI which app is being built? Like here for example: https://github.com/LedgerHQ/ledger-secure-sdk/actions/runs/12926890555/job/36051308952

Furthermore, we need to ensure that the apps built with the old CI are the same as those built with the new version.

PR #849, opened 2025-01-30
Incrementing spinner Use Case when called multiple times

nroggeman-ledger:
## Description

The goal of this PR is to implement the ability to call nbgl_useCaseSpinner() several time in a raw. It enables to "turn" the corner, for example when used in some steps of a long mono-process, preventing the automatic update.

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/lib_nbgl/src/nbgl_layout.c
+++ b/lib_nbgl/src/nbgl_layout.c
@@ -450,10 +450,13 @@ static void spinnerTickerCallback(void)
     while (i < layout->container->nbChildren) {
         if (layout->container->children[i]->type == CONTAINER) {
             nbgl_container_t *container = (nbgl_container_t *) layout->container->children[i];
-            if (container->children[0]->type == SPINNER) {
+            if (container->nbChildren && (container->children[0]->type == SPINNER)) {
                 spinner = (nbgl_spinner_t *) container->children[0];
                 spinner->position++;
-                spinner->position &= 3;  // modulo 4
+                // there are only NB_SPINNER_POSITIONS positions
+                if (spinner->position == NB_SPINNER_POSITIONS) {
+                    spinner->position = 0;
+                }
                 nbgl_objDraw((nbgl_obj_t *) spinner);
                 nbgl_refreshSpecial(BLACK_AND_WHITE_FAST_REFRESH);
                 return;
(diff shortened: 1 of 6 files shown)


lib_nbgl/src/nbgl_layout.c
```diff
@@ -3591,6 +3598,84 @@ int nbgl_layoutAddSpinner(nbgl_layout_t *layout, const char *text, const char *s
...
+
+        if (container->nbChildren != 3) {
+            return -1;
+        }
+        subTextArea = (nbgl_text_area_t *) container->children[2];
+        if (subTextArea->obj.type != TEXT_AREA) {
+            return -1;
+        }
+        // if text is different, redraw
+        if (!strcmp(subTextArea->text, text)) {
```
cedelavergne-ledger, line 3669: The comparisson should be `subText`, no?

Also, why using `strcmp` here, whereas there is a comment above (line3651) saying it cashes Rust SDK?

nroggeman-ledger: Oh yes, thanks :)

lib_nbgl/src/nbgl_layout.c
```diff
@@ -450,10 +450,13 @@ static void spinnerTickerCallback(void)
     while (i < layout->container->nbChildren) {
         if (layout->container->children[i]->type == CONTAINER) {
             nbgl_container_t *container = (nbgl_container_t *) layout->container->children[i];
-            if (container->children[0]->type == SPINNER) {
+            if (container->nbChildren && (container->children[0]->type == SPINNER)) {
                 spinner = (nbgl_spinner_t *) container->children[0];
                 spinner->position++;
-                spinner->position &= 3;  // modulo 4
+                // there are only 4 positions
+                if (spinner->position == 4) {
```
cedelavergne-ledger, line 457: This hardcoded value 4 for the number of positions is also used in `nbgl_use_case.c` (line 3896).
Should be better to have a `define`, just in case we need/want to update the mecanism later.

yogh333 (approved):

lib_nbgl/src/nbgl_layout.c
nroggeman-ledger: agree

PR #856, opened 2025-01-31
Remove useless lib_ux_sync

cedelavergne-ledger:
## Description

Remove this useless synchronous API.

No need to deprecate it because it is unused.

## Changes include

- [ ] Tests
- [ ] Documentation


tdejoigny-ledger (commented):
Shouldn't we take this opportunity to deprecate the `nbgl_useCaseRegularReview ` use case as discussed?

tdejoigny-ledger (commented):
don't forget to remove the `nbgl_useCaseRegularReview` in the doc ;-)

tdejoigny-ledger (approved):

PR #865, opened 2025-02-18
App storage functionality simplification

iartemov-ledger:
:warning: **This PR for demonstration/discussions only, not supposed to be merged. There will be another PR targetting `master`.** 

## Description

- [x] the app storage initialization is implicit, still at app boot
- [ ]  do we want to perform the initialization at first use ? 
- [x] `HAVE_APP_STORAGE`, `APP_STORAGE_SIZE`, `HAVE_APP_STORAGE_PROP_SETTINGS` and `HAVE_APP_STORAGE_PROP_DATA` have to be define in app's Makefile
- [x] the variable that points to the NVM is in a special linker section
- [ ] the user can see directly only the app data part 
    - [ ] `struct_version` is the header version (the app data versioning has to be done explicitly in the app) 
    - [ ] `size` represents the app data size and grows with writes 
- [x]  the API is simplified (3 setters, 4 getters)
- [x] Stax
- [ ] Flex
- [ ] NanoSP
- [ ] NanoX
- [ ] NanoS ? 
- [ ] discuss/remove `TODOs` 

Corresponding `app-boilerplate` change:  ?

## Changes include

- [ ] Tests
- [ ] Documentation

## Breaking changes

See above :arrow_up:


lib_standard_app/app_storage.c
```diff
@@ -20,87 +20,138 @@
...
+void app_storage_init(void)
+{
+    if (app_storage_is_initalized()) {
+        return;
+    }
+
     app_storage_header_t header = {0};
 
-    memcpy(header.tag, (void *) "NVRA", 4);
+    // TODO: should we erase all the area ?
```
iartemov-ledger, line 60: Comments from @jarevalo-ledger: "Bolos ensures the wiping out of the NVM (app storage area) at the start of each application".

include/app_storage.h
```diff
@@ -29,53 +29,43 @@
...
 
-#endif  // HAVE_BOLOS
+/* Getter for the application data */
+/* XXX: It is more practical finally to give the direct read access to the NVM for the app */
+const void *app_storage_get(void);
+
+/* Setters */
+int32_t app_storage_pwrite(const void *buf, uint32_t nbyte, uint32_t offset);
+void    app_storage_set_data_version(uint32_t data_version);
+void    app_storage_increment_data_version(void);
```
iartemov-ledger, line 69: Do we need `app_storage_erase()` to be able to dynamically reduce app storage size ?
To note that from SDK PoV we'll be able to play only if the part at the end of storage.

include/app_storage.h
```diff
@@ -29,53 +29,43 @@
...
-uint16_t app_storage_get_struct_version(void);
 uint16_t app_storage_get_properties(void);
 uint32_t app_storage_get_data_version(void);
-bool     app_storage_is_initalized(void);
-void     app_storage_set_data_version(uint32_t data_version);
 
-#endif  // HAVE_BOLOS
+/* Getter for the application data */
+/* XXX: It is more practical finally to give the direct read access to the NVM for the app */
+const void *app_storage_get(void);
```
iartemov-ledger, line 64: @jarevalo-ledger >> to reconsider `app_storage_pread()`:
+ more clean and consistent with `write()`
+/- the apps would need to keep in RAM most useful parts of its permanent data  - a bit different approach that may even speed up
- may require more stack RAM (e.g. to read a 256 bytes SK slot in a function)
- it will be always a copy (now its a direct read - but to a RAM variable )?

include/app_storage.h
```diff
@@ -29,53 +29,43 @@
...
- * @brief To be used by function accessing application storage data (not N_storage_real)
- *
- */
-#define N_app_storage (*(volatile app_storage_t *) PIC(&N_app_storage_real))
-
-void     app_storage_init(uint32_t data_version);
+/* Getters for system information */
 uint32_t app_storage_get_size(void);
-uint16_t app_storage_get_struct_version(void);
 uint16_t app_storage_get_properties(void);
```
iartemov-ledger, line 59: @jarevalo-ledger , @iartemov-ledger >> to remove the fucntion (at least so far) as:
* the app knows its properties
* the structure should be handled through the version field 
?

include/app_storage.h
```diff
@@ -29,53 +29,43 @@
...
  * @brief Structure defining the header of application storage header
  *
  */
 typedef struct app_storage_header_s {
-    char     tag[4];          ///< ['N','V','R','A'] array, when properly initialized
+    char     tag[APP_STORAGE_TAG_LEN];  ///< ['N','V','R','A'] array, when properly initialized
     uint32_t size;            ///< size in bytes of the data (app_storage_data_t structure)
-    uint16_t struct_version;  ///< version of the structure of data (to be set once at first
-                              ///< application start-up)
+    uint16_t struct_version;  ///< version of this structure (for OS)
```
iartemov-ledger, line 48: @jarevalo-ledger >> to rename to simply to `version`
@iartemov-ledger >> the idea of this step is not to break Bolos build

lib_standard_app/app_storage.c
```diff
@@ -1,5 +1,5 @@
 /*****************************************************************************
- *   (c) 2024 Ledger SAS.
+ *   (c) 2024-2025 Ledger SAS.
```
iartemov-ledger, line 2: @jarevalo-ledger >> Some useless casts to remove ...

Makefile.standard_app
```diff
@@ -53,6 +53,9 @@ endif
 ifeq ($(ENABLE_APP_STORAGE), 1)
     HAVE_APP_STORAGE = 1
     DEFINES += HAVE_APP_STORAGE
+    DEFINES += APP_STORAGE_SIZE=$(APP_STORAGE_SIZE)
```
iartemov-ledger, line 56: @jarevalo-ledger >> should we allow an app to omit the `APP_STORAGE_SIZE` and in this case a fdefault size, equla to page will be used ?

target/stax/script.ld
```diff
@@ -69,7 +69,11 @@ SECTIONS
 
     _nvram_data = .;
 
+    /* App storage */
+    *(.storage_section)
+
     /* NVM data (ex-filesystem) */
+    /* TODO: should we align the start of N_ variables on page size ? */
```
iartemov-ledger, line 76: Linked to https://github.com/LedgerHQ/ledger-secure-sdk/pull/865/files#r1961794237

target/stax/script.ld
```diff
@@ -69,7 +69,11 @@ SECTIONS
 
     _nvram_data = .;
 
+    /* App storage */
+    *(.storage_section)
```
iartemov-ledger, line 73: @jarevalo-ledger >> to add preprocessor checking for `HAVE_APP_STORAGE`.

lib_standard_app/app_storage.c
```diff
@@ -20,87 +20,138 @@
 #include "os_nvm.h"
 #include "os_pic.h"
 
-// TODO: Create new section for the linker script
-/*app_storage app_storage_real __attribute__((section(".bss.app_storage")));*/
-const app_storage_t N_app_storage_real;
+/* The storage consists of the system and the app parts */
+const uint8_t app_storage_real[sizeof(app_storage_header_t) + APP_STORAGE_SIZE]
```
iartemov-ledger, line 24: @jarevalo-ledger : to use a structure hear to simplify a bit the code.

lib_standard_app/app_storage.c
```diff
@@ -20,87 +20,138 @@
...
+    __attribute__((section(".storage_section")));
+#define app_storage_pic ((volatile uint8_t *) PIC(&app_storage_real))
```
iartemov-ledger, line 26: @jarevalo-ledger  To remove `pic`.

Makefile.standard_app
iartemov-ledger: It is done for the maximum page sizer between devices (MAX(256, 5123) minus 32 bytes for the system header with margin (now it is 16 bytes).

include/app_storage.h
iartemov-ledger: Let's leave it as it is already here ;)

include/app_storage.h
iartemov-ledger: FYI Neither `app-boilerplate` nor `app-security-key` remove some storage data today (in `app-security-key` there is an erase, but just only to add another data to the very same slot).

lib_standard_app/app_storage.c
iartemov-ledger: Some fixed

target/stax/script.ld
iartemov-ledger: We do not run processor against the .ld files. 
If no variable is declared in `storage_section` it will not occupy anything.

lib_standard_app/app_storage.c
iartemov-ledger: Cool!

include/app_storage.h
iartemov-ledger: Done

lib_standard_app/app_storage.c
iartemov-ledger: Done.

lib_standard_app/app_storage.c
iartemov-ledger: Done.

include/app_storage.h
```diff
@@ -29,53 +29,44 @@
...
  */
-typedef struct app_storage_header_s {
-    char     tag[4];          ///< ['N','V','R','A'] array, when properly initialized
+typedef struct __attribute__((packed)) app_storage_header_s {
+    char     tag[APP_STORAGE_TAG_LEN];  ///< ['N','V','R','A'] array, when properly initialized
     uint32_t size;            ///< size in bytes of the data (app_storage_data_t structure)
-    uint16_t struct_version;  ///< version of the structure of data (to be set once at first
-                              ///< application start-up)
+    uint16_t struct_version;  ///< version of this structure (for OS)
     uint16_t properties;      ///< used as a bitfield to set properties, like: contains settings,
```
cedelavergne-ledger, line 49: This field is not linked to `HAVE_APP_STORAGE_PROP_SETTINGS`?
Same for data?

include/app_storage.h
```diff
@@ -29,53 +29,44 @@
...
 uint32_t app_storage_get_data_version(void);
-bool     app_storage_is_initalized(void);
-void     app_storage_set_data_version(uint32_t data_version);
 
-#endif  // HAVE_BOLOS
+/* Reads app storage data */
+int32_t app_storage_pread(void *buf, uint32_t nbyte, uint32_t offset);
+
+/* Writes app storage data */
+int32_t app_storage_pwrite(const void *buf, uint32_t nbyte, uint32_t offset);
```
cedelavergne-ledger, line 66: I am just surprised by the naming: why this `pwrite`, and not simply `write`? We only have a single write api, no?

lib_standard_app/app_storage.c
```diff
@@ -20,87 +20,136 @@
...
-/*app_storage app_storage_real __attribute__((section(".bss.app_storage")));*/
-const app_storage_t N_app_storage_real;
+/* The storage consists of the system and the app parts */
+typedef struct __attribute__((packed)) app_storage_s {
+    app_storage_header_t header;
+    uint8_t              data[APP_STORAGE_SIZE];
+} app_storage_t;
+
+const app_storage_t app_storage_real __attribute__((section(".storage_section")));
+#define as (*(volatile app_storage_t *) PIC(&app_storage_real))
```
cedelavergne-ledger, line 30: as is not a good name... What does it mean?
Could we rather have something like `storage`, or `app_storage`?

lib_standard_app/app_storage.c
```diff
@@ -20,87 +20,136 @@
...
-    // app_storage_data.h
-    header.struct_version = APP_STORAGE_DATA_STRUCT_VERSION;
-    header.data_version   = data_version;
-    header.properties     = APP_STORAGE_PROPERTIES;
-    // TODO: Doing this lead to have app storage bigger than needed
-    header.size = sizeof(app_storage_data_t);
-    nvm_write((void *) &N_app_storage.header, &header, sizeof(header));
+    memcpy(&header.tag, (void *) APP_STORAGE_TAG, APP_STORAGE_TAG_LEN);
+    header.struct_version = APP_STORAGE_HEADER_STRUCT_VERSION;
+    header.data_version   = 1;
```
cedelavergne-ledger, line 61: Shouldn't it be an App information, given in parameter?

lib_standard_app/app_storage.c
```diff
@@ -20,87 +20,136 @@
...
+    header.properties = ((HAVE_APP_STORAGE_PROP_SETTINGS << 0) | (HAVE_APP_STORAGE_PROP_DATA << 1));
+    header.size       = 0;
```
cedelavergne-ledger, line 63: useless, because of the init above, line 57

lib_standard_app/app_storage.c
```diff
@@ -20,87 +20,136 @@
...
+    }
+    return nbyte;
 }
 
 /**
- * @brief set data version of app data
+ * @brief reads application data from the storage
  */
-void app_storage_set_data_version(uint32_t data_version)
+int32_t app_storage_pread(void *buf, uint32_t nbyte, uint32_t offset)
```
cedelavergne-ledger, line 137: why `pread`, and not simply `read`?

lib_standard_app/main.c
```diff
@@ -40,6 +40,8 @@ WEAK void __attribute__((noreturn)) app_exit(void)
     os_sched_exit(-1);
 }
 
+extern void app_storage_init(void);
```
cedelavergne-ledger, line 43: Do we consider this should be defined by the App?
Would it be simpler/clearer to define a weak function, that would be overloaded by the app, to evoid an extern definition?
`WEAK void app_storage_init(void) {}`

Makefile.standard_app
```diff
@@ -53,6 +53,13 @@ endif
 ifeq ($(ENABLE_APP_STORAGE), 1)
     HAVE_APP_STORAGE = 1
     DEFINES += HAVE_APP_STORAGE
+    ifeq ($(APP_STORAGE_SIZE),)
+    # Fall back to maximum page size for all the devices
+        APP_STORAGE_SIZE := 480 # 512 - 32 bytes for the system header
```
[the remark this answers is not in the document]
iartemov-ledger: This is a proposal from @jarevalo-ledger to use a default size if no `APP_STORAGE_SIZE` is defined in app `Makefile` (with `ENABLE_APP_STORAGE` still defined).
`install_parameters` that follow up the NVM part are page aligned in the linker script. 
In other words if there is at least 1 byte app storage or 1 byte `N_` variable a full page will be consumed anyway. 
This is also about the Flash specification with different page sizes between NanoX  (256 bytes) and others (512 bytes). As the app must remain orthogonal we choose the maximum here.
Nothing prevent the app developer from declaring explicitly APP_STORAGE_SIZE with less size knowing again that a whole page will be consumed (with a small advanatage for NanoX here).

include/app_storage.h
iartemov-ledger: Indeed, this field is a bitfield to bear HAVE_APP_STORAGE_PROP_SETTINGS and/or HAVE_APP_STORAGE_PROP_DATA flags.
This is up to app to indicate what types of storage data it uses. 
But it is used only for the indicating - behind that was an idea of the Product to use this indication on LL side.

lib_standard_app/app_storage.c
iartemov-ledger: It is available to app through `app_storage_set_data_version()` or `app_storage_increment_data_version()` functions, but not directly.
We should consider it as a service provided to the apps and to the users. The Product wanted to use this counter at restore to facilitate for the user the choice of what version of the storage to use.

lib_standard_app/app_storage.c
iartemov-ledger: See https://github.com/LedgerHQ/ledger-secure-sdk/pull/865#discussion_r1971274947 please.

lib_standard_app/app_storage.c
iartemov-ledger: Fixed.

lib_standard_app/main.c
cedelavergne-ledger: Yes, of course. My point is just to avoid this extern definition in the middle of a C code.
My peronal feeling is that I prefer to declare a WEAK empty function, that sould be overloaded by the app, rather tha nan hypothetical function that should be declared outside.
Just a question of point of view :wink:

iartemov-ledger: There should be a misunderstanding ;) 
It is not weak, it has a specific implementation (https://github.com/LedgerHQ/ledger-secure-sdk/blob/e04d6b51763f973e219e92a8a3d278be262097c6/lib_standard_app/app_storage.c#L51) , and it must be called implicit before apps starts.  
A solution to avoid `extern` here would be to introduce another header file, for internal SDK consumption only and to declare there this `app_storage_init()` function.

cedelavergne-ledger: The new header file is my favorite option :wink:

lib_standard_app/app_storage.c
```diff
@@ -20,87 +20,136 @@
...
+int32_t app_storage_pwrite(const void *buf, uint32_t nbyte, uint32_t offset)
+{
+    /* Input parameters verification */
+    if (buf == NULL) {
+        return -1;
     }
-    if (N_app_storage.header.size == 0) {
-        return false;
+
+    uint32_t size = offset + nbyte;
```
jarevalo-ledger, line 119: This addition can wrap around, we should catch it.

lib_standard_app/app_storage.c
```diff
@@ -20,87 +20,136 @@
...
 {
-    nvm_write((void *) &N_app_storage.header.data_version,
-              (void *) &data_version,
-              sizeof(N_app_storage.header.data_version));
+    /* Input parameters verification */
+    if (buf == NULL) {
+        return -1;
+    }
+
+    uint32_t size = offset + nbyte;
```
jarevalo-ledger, line 144: Same comment regarding the integer wrap around.

lib_standard_app/app_storage.c
iartemov-ledger: Fixed with `__builtin_add_overflow()`.

lib_standard_app/app_storage.c
iartemov-ledger: Fixed with `__builtin_add_overflow()`.

include/app_storage.h
iartemov-ledger: We finally thought about a application data version conversion and that a previous version's structure can take more space. 
To allow developer erase the data and avoid backing up a garbage we introduced app_storage_reset()

lib_standard_app/main.c
iartemov-ledger: Done.

iartemov-ledger:
Closing after https://github.com/LedgerHQ/ledger-secure-sdk/pull/902 merge.

PR #871, opened 2025-02-21
Modify and enrich W3C Use Cases (and related ones)

nroggeman-ledger:
## Description

The goal of this PR is to modify and enrich W3C Use Cases. It also concern security reports displayed in these Use Cases, for example for Blind Signing.

## Changes include

- [ ] Tests
- [ ] Documentation

## Breaking changes

nbgl_useCaseReviewWithWarning() function becomes nbgl_useCaseAdvancedReview(), because this second function was only used once in Ethereum


lib_nbgl/include/nbgl_use_case.h
```diff
@@ -252,10 +254,11 @@ typedef enum {
...
 typedef struct {
-    uint32_t predefinedSet;  ///< bitfield of pre-defined warnings, to be taken in @ref
-                             ///< nbgl_warningType_t set it to 0 if not using pre-defined warnings
+    uint32_t predefinedSet;    ///< bitfield of pre-defined warnings, to be taken in @ref
+                               ///< nbgl_warningType_t set it to 0 if not using pre-defined warnings
+    const char *dAppProvider;  ///< name of the dApp provider, used in some strings
+    const char *reportUrl;     ///< URL of the report, used in some strings
     const char *reportProvider;   ///< name of the security report provider, used in some strings
-    const char *reportUrl;        ///< URL of the report, used in some strings
     const char *providerMessage;  ///< Dedicated provider message. Default one will be used if NULL
```
cedelavergne-ledger, line 262: Maybe we'll need a second buffer, in the very specific cases `Other risk` or `Other threat`, where we have to display a different string between pre-transaction warning screen and security report screen?

lib_nbgl/src/nbgl_use_case.c
```diff
@@ -1880,34 +1934,36 @@ static void displaySecurityReport(uint32_t set)
...
                     || (reviewWithWarnCtx.warning->providerMessage == NULL)) {
                     bar.subText = securityReportItems[i].subText;
                 }
                 else {
-                    snprintf(tmpString,
-                             W3C_DESCRIPTION_MAX_LEN,
-                             "Web3 Checks found a risk:\n%s.",
-                             reviewWithWarnCtx.warning->providerMessage);
-                    bar.subText = tmpString;
+                    bar.subText = reviewWithWarnCtx.warning->providerMessage;
```
cedelavergne-ledger, line 1955: What about (here?) to handle the "Tap the top-right icon for more details" that may appear in the 1st warning page, but should be displayed in the security report?
Could we simply check is the sentence is present and not display it here?
Or should we use 2 variables
- providerMessageScreen: for the warning screen
- providerMessageReport: for the security report

And then, if both are defined, we use them on  the dedicated screen, else, usee the one defined in both screens?

nroggeman-ledger: Confirmed by Product that not necessary

lib_nbgl/include/nbgl_use_case.h
nroggeman-ledger: Confirmed by Product that not necessary

cedelavergne-ledger (approved):

PR #879, opened 2025-02-25
No UI/UX when calling an app in lib mode

yogh333:
## Description

When calling an app as a lib, using NBGL or BAGL is not allowed and can make the app crash.

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/lib_standard_app/main.c
+++ b/lib_standard_app/main.c
@@ -128,10 +128,6 @@ WEAK void library_app_main(libargs_t *args)
 
                         common_app_init();
 
-#ifdef HAVE_NBGL
-                        nbgl_useCaseSpinner("Signing");
-#endif  // HAVE_NBGL
-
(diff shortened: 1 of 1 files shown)


cedelavergne-ledger (approved):

lib_standard_app/main.c
```diff
@@ -128,10 +128,6 @@ WEAK void library_app_main(libargs_t *args)
 
                         common_app_init();
 
-#ifdef HAVE_NBGL
```
cedelavergne-ledger, line 131: I would rather suggest to change the flag by `SCREEN_SIZE_WALLET` to keep the existing behavior

yogh333: Even on Nano devices supporting NBGL, we could display the spinner, so I would say that this PR is irrelevant.

PR #899, opened 2025-03-11
Use reusable Rust SDK workflow to build all Rust apps for every PR

yogh333:


.github/workflows/build_rust_sdk.yml
tdejoigny-ledger, line 1: could be useful to keep the rust SDK build in the CI ? Wdyt ?

.github/workflows/build_all_rust_apps.yml
```diff
@@ -0,0 +1,20 @@
+name: Build all Rust apps
+
+on:
+  workflow_dispatch:
```
tdejoigny-ledger, line 4: I'm not able to use the "workflow_dispatch" feature

tdejoigny-ledger: it's working for the CI for the C app builds

.github/workflows/build_rust_sdk.yml
yogh333: Well the Rust SDK is systematically built when building a Rust app, so it could be redundant...

yogh333: Ok I have re-enabled it

.github/workflows/build_all_rust_apps.yml
yogh333: Well I think it will work when the PR will be merged, I have had the same issue for Rust SDK: as long as the new workflow was only in the PR, despite `workflow_dispatch`, I was not able to trigger it manually.

PR #930, opened 2025-04-16
Update Nano NBGL for new version of App SDK

nroggeman-ledger:
## Description

The goal of this PR is to support the new version of App SDK for Nano products, as defined in https://www.figma.com/design/btJTlsP9RaT0xlBFk3bd7Z/02.-App-SDK?node-id=1-98&p=f&t=oegRyLVhjvdyEruY-0

Be careful, this PR needs an OS on API level 23 to work

## Changes include

- [ ] Tests
- [ ] Documentation


lib_nbgl/src/nbgl_use_case_nanos.c
```diff
@@ -758,22 +1045,19 @@ static void displayReviewPage(nbgl_stepPosition_t pos)
...
         else if (context.currentPage == rejectIndex) {
             // Reject page
-            icon                 = &C_icon_crossmark;
-            text                 = "Reject";
-            context.stepCallback = onReviewReject;
+            getLastPageInfo(false, &icon, &text);
         }
     }
     else if (context.currentPage < reviewPages) {
         if (context.currentPage == warnIndex) {
```
cedelavergne-ledger, line 1056: `warnIndex` is no more initialized (or at least to `255`, which is no more a valid page number)!
In previous version, it was initlialized to `0` only when present, i.e. in Blind Signing mode

lib_nbgl/src/nbgl_use_case_nanos.c
```diff
@@ -795,45 +1079,43 @@ static void displayReviewPage(nbgl_stepPosition_t pos)
...
+    uint8_t                       reviewPages = 0;
+    uint8_t                       warnIndex   = 255;
+    uint8_t                       titleIndex  = 255;
+    uint8_t                       subIndex    = 255;
+    const nbgl_contentValueExt_t *extension   = NULL;
 
     context.stepCallback = NULL;
     switch (context.type) {
         case STREAMING_START_REVIEW_USE_CASE:
-        case STREAMING_BLIND_SIGN_START_REVIEW_USE_CASE:
```
cedelavergne-ledger, line 819: Same comment. `warnIndex` doesn't have any more a valid init value.
This screen is only present for Blind Signing (waiting for Tx Check :wink: )

lib_nbgl/src/nbgl_use_case_nanos.c
```diff
@@ -868,43 +1150,43 @@ static void displayStreamingReviewPage(nbgl_stepPosition_t pos)
...
                 }
-                nbgl_stepDrawText(directions,
-                                  buttonSkipCallback,
-                                  NULL,
-                                  "Press right to continue message.\nDouble-press to skip",
-                                  NULL,
-                                  REGULAR_INFO,
-                                  false);
+                info.icon  = &C_Information_circle_14px;
+                info.text1 = "Press right button to continue message or \bpress both to skip\b";
```
cedelavergne-ledger, line 1159: I have not checked, but does it really fit in the screen size?

lib_nbgl/src/nbgl_use_case_nanos.c
```diff
@@ -1095,10 +1383,8 @@ static void startUseCaseSettingsAtPage(uint8_t initSettingPage)
     nbgl_content_t        content   = {0};
     const nbgl_content_t *p_content = NULL;
 
-    if (context.type == 0) {
-        // Not yet init, it is not a GENERIC_SETTINGS
-        context.type = SETTINGS_USE_CASE;
-    }
+    context.type = SETTINGS_USE_CASE;
```
cedelavergne-ledger, line 1386: Removing this `if` means, if we call this function for a _USE_CASE_GENERIC_SETTINGS_, we'll loose the information, and switch back to a default _USE_CASE_SETTING_ :thinking:

lib_nbgl/src/nbgl_use_case_nanos.c
```diff
@@ -758,22 +1045,19 @@ static void displayReviewPage(nbgl_stepPosition_t pos)
...
+            getLastPageInfo(false, &icon, &text);
         }
     }
     else if (context.currentPage < reviewPages) {
         if (context.currentPage == warnIndex) {
             // Blind Signing Warning page
-            icon = &C_icon_warning;
-            text = "Blind\nsigning";
+            icon    = &C_icon_warning;
+            text    = "Blind signing ahead";
```
cedelavergne-ledger, line 1059: Why keeping this special warning page here, and not using the new function `displayWarningStep`?
This would avoid duplicate, and also I am wondering why we should have both calling paths now?

lib_nbgl/src/nbgl_use_case_nanos.c
```diff
@@ -1865,16 +2273,15 @@ void nbgl_useCaseReviewBlindSigning(nbgl_operationType_t              operationT
...
-                  reviewSubTitle,
-                  finishTitle,
-                  choiceCallback);
+    nbgl_useCaseAdvancedReview(operationType,
+                               tagValueList,
+                               icon,
+                               reviewTitle,
+                               reviewSubTitle,
+                               finishTitle,
+                               dummy,
```
cedelavergne-ledger, line 2282: Shouldn'y we rename this, like for Stax/Flex? `dummy` was my previous naming, because useless, which is no more the case?

lib_nbgl/src/nbgl_use_case_nanos.c
```diff
@@ -2099,18 +2503,73 @@ void nbgl_useCaseReviewStreamingBlindSigningStart(nbgl_operationType_t       ope
...
+ *
+ * @param operationType type of operation (Operation, Transaction, Message)
+ * @param icon icon used on first and last review page
+ * @param reviewTitle string used in the first review page
+ * @param reviewSubTitle string to set under reviewTitle (can be NULL)
+ * @param warning structure to build the initial warning page (cannot be NULL)
+ * @param choiceCallback callback called when more operation data are needed (param is true) or
+ * operation is rejected (param is false)
+ */
+void nbgl_useCaseAdvancedReviewStreamingStart(nbgl_operationType_t       operationType,
```
cedelavergne-ledger, line 2525: Don't you have lots of duplications here with `nbgl_useCaseAdvancedReview`?
My feeling is that we could factorize, just updating the `type` ?

lib_nbgl/src/nbgl_use_case_nanos.c
```diff
@@ -476,24 +550,69 @@ static void drawStep(nbgl_stepPosition_t        pos,
...
+                           nbgl_stepButtonCallback_t onActionCallback,
+                           bool                      modal)
+{
+    nbgl_screenTickerConfiguration_t *p_ticker = NULL;
+    nbgl_screenTickerConfiguration_t  ticker   = {
+           .tickerCallback  = PIC(statusTickerCallback),
+           .tickerIntervale = 0,    // not periodic
+           .tickerValue     = 3000  // 3 seconds
+    };
+    nbgl_layoutSwitch_t switchInfo;
```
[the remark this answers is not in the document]
nroggeman-ledger: I'm already filling the only 3 used fields so useless

lib_nbgl/src/nbgl_use_case_nanos.c
nroggeman-ledger: I've generated screenshots for all these flows (it will be in another PR)

lib_nbgl/src/nbgl_use_case_nanos.c
nroggeman-ledger: I agree, will remove that

lib_nbgl/src/nbgl_use_case_nanos.c
nroggeman-ledger: no, the tipbox is still useless

lib_nbgl/src/nbgl_use_case_nanos.c
nroggeman-ledger: It was not working for the normal Home & Settings with the current version, I will try to fix that

cedelavergne-ledger (approved):

PR #933, opened 2025-04-18
Update Nano NBGL for new version of App SDK

nroggeman-ledger:
## Description

The goal of this PR is to support the new version of App SDK for Nano products, as defined in https://www.figma.com/design/btJTlsP9RaT0xlBFk3bd7Z/02.-App-SDK?node-id=1-98&p=f&t=oegRyLVhjvdyEruY-0

## Changes include

- [ ] Tests
- [ ] Documentation


cedelavergne-ledger (approved):

tools/ttf2inc.py
dmorais-ledger, line 1: I guess this is too late because that PR is already merged, but I think there is a problem here: Stax & Flex also use 1 BPP fonts, so the constraint on Y should remain even when self.bpp == 1.
Also, I don't understand why the lines taking care of removing the transparent pixels have been removed, as it was a way to remove unnecessary data.

PR #943, opened 2025-04-29
Improvements in SDK Fuzzing

Gustavo-Jodar:
## Description

Add ClusterFuzzLite to Ledger-Secure-SDK and three fuzzers:
 - `fuzzer_bip32`
 - `fuzzer_lib_alloc`
 - `fuzzer_nfc_ndef`

Update documentation accordingly .

## Changes include

- [X] Tests
- [X] Documentation

## Breaking changes

Changed the function `os_ndef_to_string` in `lib_nfc/src/nfc_ndef.c`

### DIFF
--- a/.clusterfuzzlite/Dockerfile
+++ b/.clusterfuzzlite/Dockerfile
@@ -1,0 +1,13 @@
+FROM ghcr.io/ledgerhq/ledger-app-builder/ledger-app-builder-lite:latest AS LITE_BUILDER
+
+# Base image with clang toolchain
+FROM gcr.io/oss-fuzz-base/base-builder:v1
+
+# Copy the project's source code.
+COPY . $SRC/ledger-secure-sdk
+
+# Working directory for build.sh
+WORKDIR $SRC/ledger-secure-sdk
+
+# Copy build.sh into $SRC dir.
+COPY ./.clusterfuzzlite/build.sh $SRC/
--- a/.clusterfuzzlite/build.sh
+++ b/.clusterfuzzlite/build.sh
@@ -3,7 +3,8 @@
 # build fuzzers
 
 pushd fuzzing
-cmake -B build
+rm -rf build/*
+cmake -B build -S . -DCMAKE_C_COMPILER=/usr/bin/clang -DSANITIZER=address
 make -C build
-mv ./build/fuzz_bip32 "${OUT}"
+mv ./build/fuzz_alloc "${OUT}"
 popd
--- a/.clusterfuzzlite/build.sh
+++ b/.clusterfuzzlite/build.sh
@@ -6,5 +6,5 @@ pushd fuzzing
 rm -rf build/*
 cmake -B build -S . -DCMAKE_C_COMPILER=/usr/bin/clang -DSANITIZER=address
 make -C build
-mv ./build/fuzz_alloc "${OUT}"
+mv ./build/fuzz_* "${OUT}"
 popd
--- a/.clusterfuzzlite/build.sh
+++ b/.clusterfuzzlite/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash -eu
+
+# build fuzzers
+
+pushd fuzzing
+cmake -B build
+make -C build
+mv ./build/fuzz_bip32 "${OUT}"
+popd
--- a/fuzzing/CMakeLists.txt
+++ b/fuzzing/CMakeLists.txt
@@ -12,14 +12,17 @@ set(CMAKE_BUILD_TYPE "Debug")
 
 set(CMAKE_C_STANDARD 11)
 set(CMAKE_C_FLAGS
-    "${CMAKE_C_FLAGS} -Wall -Wextra -pedantic -DFUZZ -g -O0 -fsanitize=fuzzer,address,undefined"
+    "${CMAKE_C_FLAGS} -Wall -Wextra -pedantic -DFUZZ -g -O0 -fdata-sections -ffunction-sections -funsigned-char -fshort-enums -fsanitize=fuzzer,address,undefined -fprofile-instr-generate -fcoverage-mapping"
 )
 
 include_directories(
-  ../lib_standard_app/
-  ../qrcode/include/
   mock/
+  ../lib_standard_app/
+  ../include/
+  ../qrcode/include/
 )
+
+add_definitions(-DHAVE_NFC -DHAVE_NDEF_SUPPORT)
 
 add_library(base58 SHARED ../../lib_standard_app/base58.c)
 add_library(bip32 SHARED ../../lib_standard_app/bip32.c)
--- a/fuzzing/CMakeLists.txt
+++ b/fuzzing/CMakeLists.txt
@@ -1,16 +1,16 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.14)
 
 set(CMAKE_C_COMPILER clang)
 set(CMAKE_CPP_COMPILER clang)
 
-if(${CMAKE_VERSION} VERSION_LESS 3.10)
+if(${CMAKE_VERSION} VERSION_LESS 3.14)
     cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
 endif()
 
 # project information
-project(FuzzBip32
+project(SDKFuzzer
         VERSION 1.0
-	      DESCRIPTION "Fuzzing of BIP32"
+	        DESCRIPTION "SDK Fuzzer"
         LANGUAGES C)
 
 # guard against bad build-type strings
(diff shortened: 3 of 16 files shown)


.clusterfuzzlite/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ghcr.io/ledgerhq/ledger-app-builder/ledger-app-builder-lite:latest AS LITE_BUILDER
```
0pendev, line 1: Maybe I'm missing something but when/how is LITE_BUILDER used here?

fuzzing/extra/nfc_ndef.cmake
```diff
@@ -0,0 +1,19 @@
+include_guard()
+
+add_definitions(-DHAVE_NFC -DHAVE_NDEF_SUPPORT)
```
0pendev, line 3: Let's not add_definitions to the project.
Either use target_compile_definitions to add it only to nfc_ndef or make an interface target 'common' that you link your targets agains.

fuzzing/mock/exceptions.h
```diff
@@ -0,0 +1,3 @@
+#pragma once
+
+#define THROW exit
```
0pendev, line 3: Are we sure this is the behavior we want? @bboilot-ledger

fuzzing/mock/os_task.c
```diff
@@ -4,3 +4,13 @@ void os_sched_exit(__attribute__((unused)) bolos_task_status_t exit_code)
 {
     return;
 }
+
+void os_longjmp(unsigned int exception)
+{
+#ifdef HAVE_DEBUG_THROWS
```
0pendev, line 10: Let's try and keep it as simple as possible on the mocks.
Either we want the DEBUG_THROWS on the mocks or we don't.
I don't see the use of handling both.

@bboilot-ledger what do you think?

fuzzing/README.md
```diff
@@ -24,7 +24,7 @@ Once in the container, go into the `fuzzing` folder to compile the fuzzer:
 cd fuzzing
```
0pendev, line 24: Could be an opportunity to use cmake only commands (could ease the use of alternative engines like Ninja instead of Makefile).
The `cmake --build` command also supports targets specification with `--target`
Also, why are we giving an absolute path to clang?
```suggestion
cmake -S fuzzing -B build -DCMAKE_C_COMPILER=clang -DSANITIZER=address
cmake --build build
```

lib_nfc/src/nfc_ndef.c
```diff
@@ -168,34 +168,36 @@ uint8_t os_parse_ndef(uint8_t *in_buffer, ndef_struct_t *parsed)
  */
 uint16_t os_ndef_to_string(ndef_struct_t *ndef_message, char *out_string)
```
0pendev, line 169: @bboilot-ledger do you think we should make a dedicated PR for this fiz?

fuzzing/CMakeLists.txt
```diff
@@ -1,38 +1,59 @@
-cmake_minimum_required(VERSION 3.10)
-
-project(Fuzzer
-        VERSION 1.0
-	      DESCRIPTION "SDK Fuzzer"
-        LANGUAGES C)
+cmake_minimum_required(VERSION 3.14)
 
 set(CMAKE_C_COMPILER clang)
```
0pendev, line 3: Do we need those lines? The README already set clang as an argument.
@bboilot-ledger what would be easier for devs?

fuzzing/CMakeLists.txt
```diff
@@ -1,38 +1,59 @@
...
+	        DESCRIPTION "SDK Fuzzer"
+        LANGUAGES C)
+
+# guard against bad build-type strings
+if (NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE "Debug")
+endif()
+
+
+if (NOT CMAKE_C_COMPILER_ID MATCHES "Clang")
```
0pendev, line 22: If we keep the previous compiler definitions we could remove this if

fuzzing/CMakeLists.txt
```diff
@@ -1,38 +1,59 @@
...
+    message(FATAL_ERROR "Fuzzer needs to be built with Clang")
+endif()
+
+# default fuzz device target
+if (NOT TARGET_DEVICE)
```
0pendev, line 27: Do we even use TARGET_DEVICE?

fuzzing/CMakeLists.txt
```diff
@@ -1,38 +1,59 @@
...
+    set(COMPILATION_FLAGS ${COMPILATION_FLAGS} -fsanitize=fuzzer,memory,undefined -fsanitize-memory-track-origins -fsanitize=fuzzer-no-link)
+  else()
+    message(FATAL_ERROR "Unknown sanitizer type. It must be set to `address` or `memory`.")
+  endif()
+else()
+	set(COMPILATION_FLAGS "$ENV{LIB_FUZZING_ENGINE} $ENV{CFLAGS}")
+  separate_arguments(COMPILATION_FLAGS)
+endif()
+
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
0pendev, line 52: Let's remove this line.
This should be set when calling cmake and is already set automatically by most Editors
```suggestion
```

0pendev (changes requested):
A nice first iteration, should be easier to add more targets in the future

fuzzing/mock/os_task.c
bboilot-ledger: Let's remove it!

lib_nfc/src/nfc_ndef.c
bboilot-ledger: The fuzzer will fail on every PR if we don't fix this code. Also, I don't see any app depending on this lib so it seems to be the best solution IMO. But will definitely ask the embedded team before merging the PR.

.clusterfuzzlite/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -eu
+
+# build fuzzers
+
+pushd fuzzing
+rm -rf build/*
+cmake -B build -S . -DCMAKE_C_COMPILER=/usr/bin/clang -DSANITIZER=address
```
bboilot-ledger, line 7: `-DSANITIZER=address` should not be required there because this condition `if (NOT DEFINED ENV{LIB_FUZZING_ENGINE})` will be false (Clusterfuzzlite will actually configure the compilation flags all by itself).

fuzzing/extra/Bip32.cmake
```diff
@@ -0,0 +1,15 @@
+include_guard()
```
bboilot-ledger, line 1: Should we name this file `bip32.cmake` to keep the same naming convention.

fuzzing/fuzzer_alloc.c
```diff
@@ -0,0 +1,145 @@
...
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+#ifdef DEBUG_CRASH
+    printf("[INIT] size = %lu\n", HEAP_SIZE);
+#endif
+    memset(heap, 0, HEAP_SIZE);
+    memset(allocs, 0, sizeof(allocs));
+    mem_ctx_t *allocator = mem_init(heap, HEAP_SIZE);  // We need a check on size else the mem_alloc
+                                                       // fails afterwards. For example for 0xffff
```
bboilot-ledger, line 106: This comment can be removed, my bad.

fuzzing/CMakeLists.txt
Gustavo-Jodar: Not now, but in next versions of the fuzzing framework (already calling it a framework) it will be necessary since some syscalls mock functions depend on definitions in /target.${TARGET_DEVICE}/include , but we can remove it for now.

fuzzing/README.md
Gustavo-Jodar: Should it be better to build the fuzz in /build or /fuzzing/build/ ? Thinking about what could be better for the dev

.clusterfuzzlite/Dockerfile
Gustavo-Jodar: You are right, I had taken the Dockerfile of boilerplate and added some changes to it forgetting to remove this image.

fuzzing/CMakeLists.txt
Gustavo-Jodar: I removed:
```set(CMAKE_C_COMPILER clang)```
```set(CMAKE_CPP_COMPILER clang)```
, so we keep this if?

File: fuzzing/extra/Bip32.cmake (hunk 12 above)
Comment (Gustavo-Jodar) [reply fuzzing/extra/Bip32.cmake:1]: I will change it.

File: fuzzing/README.md (hunk 5 above)
Comment (bboilot-ledger) [reply fuzzing/README.md:24]: Let's go with:
```
cmake -S fuzzing -B fuzzing/build -DCMAKE_C_COMPILER=clang -DSANITIZER=address
cmake --build fuzzing/build
```
You can resolve the conversation afterward.

PR #950, opened 2025-04-30
Make improvements

cedelavergne-ledger:
- Add generic standard app main header file (`main_std_app.h`) to be used by the apps
- Add rules allowing to compile Apps based on the pre-defined use-cases. 
  For example, in `app-ethereum`: `make dbg_use_test_keys`, instead of giving the whole set of flags

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -99,4 +99,17 @@ OBJECTS_DIR += $(sort $(dir $(OBJECT_FILES)))
 DEPEND_FILES = $(subst $(OBJ_DIR), $(DEP_DIR), $(addsuffix .d, $(basename $(OBJECT_FILES))))
 DEPEND_DIR += $(sort $(dir $(DEPEND_FILES)))
 
+# Path to the TOML file
+APP_MANIFEST_FILE := $(APP_DIR)/ledger_app.toml
+# If the TOML file exists, retrieve the use-cases
+ifneq ($(wildcard $(APP_MANIFEST_FILE)),)
+  # Retrieve the use-cases from the TOML file
+  APP_USE_CASES := $(shell ledger-manifest "$(APP_MANIFEST_FILE)" -ouc -j | jq -r 'keys | join(" ")')
+
+  # Function to extract flags for a specific use case
+  define get_flags_for_use_case
+    $(shell ledger-manifest "$(APP_MANIFEST_FILE)" -ouc $1)
+  endef
+endif
+
 include $(BOLOS_SDK)/Makefile.rules_generic
--- a/Makefile.rules_generic
+++ b/Makefile.rules_generic
@@ -75,6 +75,13 @@ prepare:
 
 $(BUILD_DEPENDENCIES): prepare
 
+# Define a generic rule for each use-case
+ifneq ($(APP_USE_CASES),)
+$(APP_USE_CASES):
+	@echo "Building with '$@' configuration..."
+	@$(MAKE) --no-print-directory $(call get_flags_for_use_case,$@)
+endif
+
 BIN_TARGETS := bin/app.elf bin/app.apdu bin/app.sha256 bin/app.hex
 DBG_TARGETS := debug/app.map debug/app.asm
 
--- a/lib_standard_app/main_std_app.h
+++ b/lib_standard_app/main_std_app.h
@@ -1,0 +1,31 @@
+#pragma once
+
+#include "macros.h"
+
+/**
+ * Exit the application and go back to the dashboard.
+ */
+void __attribute__((noreturn)) app_exit(void);
+
+/**
+ * Common init of the App.
+ *  - UX
+ *  - io
+ *  - USB
+ *  - BLE
+ */
+WEAK void common_app_init(void);
+
+/**
+ * Main entry point of the App, in standalone mode.
+ * It calls (inside a try/catch):
+ *  - common_app_init
+ *  - app_main (should be defined in the App)
+ */
+WEAK void standalone_app_main(void);
+
+/**
+ * Main entry point of the App, in library mode.
+ * It calls the swap handlers (inside a try/catch)
+ */
+WEAK void library_app_main(void);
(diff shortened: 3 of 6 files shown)


Makefile.rules
```diff
@@ -99,4 +99,17 @@ OBJECTS_DIR += $(sort $(dir $(OBJECT_FILES)))
 DEPEND_FILES = $(subst $(OBJ_DIR), $(DEP_DIR), $(addsuffix .d, $(basename $(OBJECT_FILES))))
 DEPEND_DIR += $(sort $(dir $(DEPEND_FILES)))
 
+# Path to the TOML file
+APP_MANIFEST_FILE := $(APP_DIR)/ledger_app.toml
+
+ifneq ($(wildcard $(APP_MANIFEST_FILE)),)
+  # Retrieve the use-cases from the TOML file
+  APP_USE_CASES := $(shell python3 $(BOLOS_SDK)/extract_use_cases.py -m "$(APP_MANIFEST_FILE)" -c names)
```
fbeutin-ledger, line 107: I like the idea of building use cases a lot but I really don't think it's the SDK Makefile responsibility to do it..

lib_standard_app/main_std_app.h
```diff
@@ -0,0 +1,31 @@
...
+ *  - common_app_init
+ *  - app_main (should be defined in the App)
+ */
+WEAK void standalone_app_main(void);
+
+/**
+ * Main entry point of the App, in library mode.
+ * It calls the swap handlers (inside a try/catch)
+ */
+WEAK void library_app_main(void);
```
fbeutin-ledger, line 31: Why do we need this now btw?

extract_use_cases.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python3
```
fbeutin-ledger, line 1: This is very redundant with ledgered

lib_standard_app/main_std_app.h
cedelavergne-ledger: The idea is to provide the different function signature, allowing the Apps to just inclide the header, instead of addinf the function prototype everywhere they need a fonction from this file

Makefile.rules
cedelavergne-ledger: Yes and No: The SDK is the central point to build the apps. All rules, configs and flags are centralized here.
The idea also is to have something independent of other tools (should it be python, bash or whatever) and usable everywhere, like VSCode, but also CI workflows (like app-tester, reusable workflows, ...)

extract_use_cases.py
cedelavergne-ledger: You are right, but ledgered is not used that way, and outputs data in different formats, and uses printf outputs that are annoying here...

lpascal-ledger: Just glanced over the code, don't have opinion. My 2 cents:

- `ledgered` will need to be `pip install`ed, which would take a bit of time.
- once installed, it could be use directly in the shell, data can be easily retrieved with `jq` or something else
```sh
$ ledger-manifest ledger_app.toml -ou -j  | jq keys
[
  "cal_bypass",
  "dbg_cal_bypass",
  "dbg_use_test_keys",
  "use_test_keys"
]
$ ledger-manifest ledger_app.toml -ou -j  | jq '.cal_bypass'
"BYPASS_SIGNATURES=1"
```
- or, if the parsing is too complex, through a Python script (`ledgered` is first a Python library)
```python
> from ledgered.manifest import Manifest
> from pathlib import Path
> sorted(Manifest.from_path(Path("ledger_app.toml")).use_cases.cases.keys())
['cal_bypass', 'dbg_cal_bypass', 'dbg_use_test_keys', 'use_test_keys']
>  Manifest.from_path(Path("ledger_app.toml")).use_cases.cases["cal_bypass"]
 'BYPASS_SIGNATURES=1'
```

cedelavergne-ledger: Changed and based now on `ledgered`

Makefile.rules
```diff
@@ -99,4 +99,17 @@ OBJECTS_DIR += $(sort $(dir $(OBJECT_FILES)))
...
 
+# Path to the TOML file
+APP_MANIFEST_FILE := $(APP_DIR)/ledger_app.toml
+
+ifneq ($(wildcard $(APP_MANIFEST_FILE)),)
+  # Retrieve the use-cases from the TOML file
+  APP_USE_CASES := $(shell ledger-manifest "$(APP_MANIFEST_FILE)" -ouc -j | jq -r 'keys | join(" ")')
+
+  # Function to extract flags for a specific use case
+  define get_flags
```
fbeutin-ledger, line 110: get_use_case_flags

Makefile.rules
```diff
@@ -99,4 +99,17 @@ OBJECTS_DIR += $(sort $(dir $(OBJECT_FILES)))
 DEPEND_FILES = $(subst $(OBJ_DIR), $(DEP_DIR), $(addsuffix .d, $(basename $(OBJECT_FILES))))
 DEPEND_DIR += $(sort $(dir $(DEPEND_FILES)))
 
+# Path to the TOML file
+APP_MANIFEST_FILE := $(APP_DIR)/ledger_app.toml
```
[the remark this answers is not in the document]
cedelavergne-ledger: Not exactly:
the `:=` ensures an init of the variable at the begining, and not only when used. `?=` will set the variable when used if not previously set. Here, the idea is really to init at the begining.
The next statement `ifneq` allos to retrieve the use-cases only if the manifest file exists.

Makefile.rules
cedelavergne-ledger: Fixed

Makefile.rules_generic
```diff
@@ -75,6 +75,13 @@ prepare:
 
 $(BUILD_DEPENDENCIES): prepare
 
+# Define a generic rule for all use cases
```
fbeutin-ledger, line 78: Nitpick
for each use case

cedelavergne-ledger: Fixed

fbeutin-ledger (approved):

PR #952, opened 2025-05-06
Fix nbgl nano

cedelavergne-ledger:
Fix `nbgl_useCaseNavigableContent` (only the 1st content was selectable)
Fix `nbgl_useCaseHomeAndSettings`, `nbgl_useCaseGenericConfiguration` and `nbgl_useCaseGenericSettings` with `BARS_LIST` or `CHOICES_LIST`

### DIFF
--- a/lib_nbgl/include/nbgl_use_case.h
+++ b/lib_nbgl/include/nbgl_use_case.h
@@ -559,7 +559,7 @@ void nbgl_useCaseKeypadDigits(const char             *title,
                               uint8_t                 maxDigits,
                               bool                    shuffled,
                               nbgl_pinValidCallback_t validatePinCallback,
-                              nbgl_callback_t         backCallbackk);
+                              nbgl_callback_t         backCallback);
 void nbgl_useCaseKeypadPIN(const char             *title,
                            uint8_t                 minDigits,
                            uint8_t                 maxDigits,
--- a/lib_nbgl/src/nbgl_use_case_nanos.c
+++ b/lib_nbgl/src/nbgl_use_case_nanos.c
@@ -675,7 +675,8 @@ static bool buttonGenericCallback(nbgl_buttonEvent_t event, nbgl_stepPosition_t 
             if (context.stepCallback != NULL) {
                 context.stepCallback();
             }
-            else if ((context.type == CONTENT_USE_CASE)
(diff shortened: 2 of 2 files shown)


lib_nbgl/include/nbgl_use_case.h
```diff
@@ -553,6 +553,13 @@ void nbgl_useCaseKeypadPIN(const char                *title,
 #endif  // NBGL_KEYPAD
 
 #else  // HAVE_SE_TOUCH
+void nbgl_useCaseNavigableContent(const char                *title,
+                                  uint8_t                    initPage,
+                                  uint8_t                    nbPages,
+                                  nbgl_callback_t            quitCallback,
+                                  nbgl_contentCallback_t     navCallback,
```
nroggeman-ledger, line 560: don't call it navCallback

nroggeman-ledger (changes requested):

lib_nbgl/include/nbgl_use_case.h
cedelavergne-ledger: Fixed

lib_nbgl/include/nbgl_use_case.h
```diff
@@ -506,6 +499,13 @@ DEPRECATED void nbgl_useCaseSettings(const char                *settingsTitle,
                                      nbgl_callback_t            quitCallback,
                                      nbgl_navCallback_t         navCallback,
                                      nbgl_layoutTouchCallback_t controlsCallback);
+void            nbgl_useCaseNavigableContent(const char                *title,
```
iartemov-ledger, line 502: 1/ Is it really mandatory to add 2 slightly different signatures of the function for `HAVE_SE_TOUCH` and `!HAVE_SE_TOUCH` cases? Could not it be factorized ?  
2/ Attention: `nbgl_useCaseNavigableContent()`  is also used in `app-security-key` and `app-monero` ?

cedelavergne-ledger: I didn't find a way to factorize, due to the difference of callbacks prototypes and the way to get the contents.
This is due also to a complex mix of `nbgl_content_t` vs `nbgl_pageContent_t` between the Nano version and the global one.

nroggeman-ledger (approved):

PR #969, opened 2025-05-19
Fuzzing framework

Gustavo-Jodar:
## Description

New fuzzing framework for Ledger SDK.
If you want to test the framework with an app, you can have a look at the boilerplate PR https://github.com/LedgerHQ/app-boilerplate/pull/163

## Changes include

- [x] Tests
- [ ] Documentation

### DIFF
--- a/fuzzing/README.md
+++ b/fuzzing/README.md
@@ -45,9 +45,11 @@ docker run --rm -ti -v "$(realpath .):/app" ghcr.io/ledgerhq/ledger-app-builder/
 ```
 
 ```console
+export BOLOS_SDK=/app
+
 cd fuzzing # You must run it from the fuzzing folder
 
-./local_run.sh --build=1 --fuzzer=build/fuzz_bip32 --j=4 --run-fuzzer=1 --compute-coverage=1
+./local_run.sh --build=1 --fuzzer=build/fuzz_bip32 --j=4 --run-fuzzer=1 --compute-coverage=1 --BOLOS_SDK=${BOLOS_SDK}
 ```
 
 ### About local_run.sh
@@ -123,7 +125,7 @@ cmake -S . -B build -DCMAKE_C_COMPILER=clang -DSANITIZER=address -G Ninja
 cmake --build build
 ```
 
-One can still use his own modified `ledgere-secure-sdk`. If it does't contain a .target, you can pass it in the compilation
+One can still use his own modified `ledgere-secure-sdk`. If it doesn't contain a .target, you can pass it in the compilation
 parameters:
 ```console
 cmake -S . -B build -DCMAKE_C_COMPILER=clang -DSANITIZER=address -G Ninja -DTARGET=stax


fuzzing/libs/glyphs.cmake
```diff
@@ -0,0 +1,58 @@
+include_guard()
+include(${BOLOS_SDK}/fuzzing/macros/macros.cmake)
+
+find_package(Python3 REQUIRED)
+find_program(MKDIR_EXECUTABLE mkdir REQUIRED)
```
0pendev, line 5: Recently I stumbled upon an easier way to create a directory:

https://cmake.org/cmake/help/latest/command/file.html#make-directory

fuzzing/libs/lib_cxng.cmake
```diff
@@ -0,0 +1,55 @@
+include_guard()
+include(${BOLOS_SDK}/fuzzing/macros/macros.cmake)
+include(${BOLOS_SDK}/fuzzing/mock/mock.cmake)
+
+file(
+  GLOB
```
0pendev, line 6: We could remove the use of file/GLOB if we need to specify all the files anyway.
If you did this because there is some files that you did not want to include, you can do your globing and then use list(REMOVE_ITEM ) to remove what you needed to exclude

fuzzing/libs/lib_io.cmake
```diff
@@ -0,0 +1,53 @@
+include_guard()
+include(${BOLOS_SDK}/fuzzing/macros/macros.cmake)
+include(${BOLOS_SDK}/fuzzing/libs/lib_nbgl.cmake)
+include(${BOLOS_SDK}/fuzzing/libs/lib_cxng.cmake)
+include(${BOLOS_SDK}/fuzzing/libs/lib_standard_app.cmake)
+
+file(
+  GLOB
+  LIB_IO_SOURCES
+  "${BOLOS_SDK}/src/os_io_usb.c"
```
0pendev, line 10: Same comment as for cxng

fuzzing/libs/lib_io.cmake
```diff
@@ -0,0 +1,53 @@
...
+  "${BOLOS_SDK}/lib_stusb/STM32_USB_Device_Library/Class/HID/Src/usbd_hid.c"
+  "${BOLOS_SDK}/lib_stusb_impl/u2f_impl.c"
+  "${BOLOS_SDK}/lib_stusb_impl/u2f_io.c"
+  "${BOLOS_SDK}/lib_stusb_impl/usbd_impl.c"
+  "${BOLOS_SDK}/lib_ux_nbgl/ux.c")
+
+add_library(io ${LIB_IO_SOURCES})
+target_link_libraries(io PUBLIC nbgl cxng standard_app macros)
+target_compile_options(io PUBLIC ${COMPILATION_FLAGS})
+target_include_directories(io PUBLIC ${BOLOS_SDK}/include/)
```
0pendev, line 35: You can make one call to target_include_directories and give multiple directories

fuzzing/libs/lib_nbgl.cmake
```diff
@@ -0,0 +1,52 @@
...
+  "${BOLOS_SDK}/lib_nbgl/src/nbgl_use_case_nanos.c")
+
+add_library(nbgl ${LIB_NBGL_SOURCES})
+target_link_libraries(nbgl PUBLIC macros glyphs qrcode mock io)
+target_compile_options(nbgl PUBLIC ${COMPILATION_FLAGS})
+target_include_directories(nbgl PUBLIC ${LIB_NBGL_FONT_DIRS})
+target_include_directories(
+  nbgl
+  PUBLIC ${BOLOS_SDK}/fuzzing/glyphs/
+         ${BOLOS_SDK}/lib_nbgl/include/fonts/
```
0pendev, line 46: Why are they separated?

```suggestion
target_include_directories(nbgl PUBLIC
 ${LIB_NBGL_FONT_DIRS}
 ${BOLOS_SDK}/fuzzing/glyphs
 ${BOLOS_SDK}/lib_nbgl/include/fonts
```

fuzzing/libs/lib_standard_app.cmake
```diff
@@ -0,0 +1,33 @@
+include_guard()
+include(${BOLOS_SDK}/fuzzing/macros/macros.cmake)
+include(${BOLOS_SDK}/fuzzing/libs/lib_cxng.cmake)
+include(${BOLOS_SDK}/fuzzing/mock/mock.cmake)
+
+# Commented lines need os mock or depend of another function that needs a mock
+file(
+  GLOB
```
0pendev, line 8: Same comment as for cxng

fuzzing/local_run.sh
```diff
@@ -0,0 +1,153 @@
...
+    echo "  --TARGET_DEVICE=[flex|stax] Whether it is a flex or stax device (default: flex)"
+    echo "  --build=1|0                 Whether to build the project (default: 0)"
+    echo "  --compute-coverage=1|0      Whether to compute coverage after fuzzing (default: 0)"
+    echo "  --run-fuzzer=1|0            Whether to run or not the fuzzer (default: 1)"
+    echo "  --help                      Show this help message"
+    exit 0
+}
+
+function build(){
+    # TODO -> remove those lines after installation in the docker image is done by default
```
0pendev, line 24: Isn't done?

fuzzing/local_run.sh
```diff
@@ -0,0 +1,153 @@
+#!/bin/bash
```
0pendev, line 1: Please run shellcheck on this script.
It's a quick win and can prevent some issues like the lack of double quoting

fuzzing/mock/custom/custom_syscalls.c
```diff
@@ -0,0 +1,44 @@
...
+#include "ox_vss.h"
+#endif  // HAVE_VSS
+#include "os_seed.h"
+#include "ox_crc.h"
+#include "os_endorsement.h"
+#include <string.h>
+
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+
+cx_err_t cx_ecdomain_parameters_length(cx_curve_t cv, size_t *length)
```
0pendev, line 40: Is this function suposed to be here?

fuzzing/mock/custom/custom_syscalls.c
```diff
@@ -0,0 +1,44 @@
...
+#endif
+#if defined(HAVE_VSS)
+#include "ox_vss.h"
+#endif  // HAVE_VSS
+#include "os_seed.h"
+#include "ox_crc.h"
+#include "os_endorsement.h"
+#include <string.h>
+
+#pragma GCC diagnostic ignored "-Wunused-parameter"
```
0pendev, line 38: Let's not have such a pragma here. It will be applied to any file that includes this header.
If it's the behavior we need then let's consider using -Wno-unused-parameter in our compilation flags.

If it was meant for the function bellow then either add a `cx_curve_t cv __attribute__((unused))` or a simple `(void) cv`

fuzzing/mock/mock.cmake
```diff
@@ -0,0 +1,31 @@
+include_guard()
+
+include(${BOLOS_SDK}/fuzzing/macros/macros.cmake)
+include(${BOLOS_SDK}/fuzzing/libs/lib_nbgl.cmake)
+include(${BOLOS_SDK}/fuzzing/libs/lib_standard_app.cmake)
+include(${BOLOS_SDK}/fuzzing/libs/lib_cxng.cmake)
+
+# Commented lines need os mock or depend of another function that needs a mock
+file(
+  GLOB
```
0pendev, line 10: same comment as for cxng

0pendev (changes requested):
Nice iteration. Will test it but it looks fairly easy to use and documented enough for new commers.

Made a review on the CMake scripts.

fuzzing/CMakeLists.txt
```diff
@@ -1,38 +1,103 @@
...
+    FATAL_ERROR
+      "In-source builds not allowed. Please make a new directory (called a build directory) and run CMake from there. You may need to remove CMakeCache.txt. "
+  )
+endif()
+
+# compatible with ClusterFuzzLite
+if(NOT DEFINED ENV{LIB_FUZZING_ENGINE})
+  set(COMPILATION_FLAGS
+      -g
+      -O0
```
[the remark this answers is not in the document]
Gustavo-Jodar: From the compile_commands.json I was just using the macros, should I also add the flags generated by the compilation of the app?

fuzzing/mock/mock.cmake
```diff
@@ -0,0 +1,31 @@
+include_guard()
```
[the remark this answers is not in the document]
Gustavo-Jodar: I put a very quick mention in the ```README.md```, but I can put it in other place as well, like in a README.md for the mock or in a comment in a file. What you think?

fuzzing/README.md
```diff
@@ -4,30 +4,72 @@
...
+| `--fuzzer` | `PATH` | **Required**. Path to the fuzzer binary |
+| `--compute-coverage` | `bool` | **Optional**. Whether to compute coverage after fuzzing (default: 0) |
+| `--run-fuzzer` | `bool` | **Optional**. Whether to run or not the fuzzer (default: 1) |
+| `--help` |  | **Optional**. Dsplay help message |
+
+
 ### Compilation
 
 Once in the container, go into the `fuzzing` folder to compile the fuzzer:
+Make sure that the packages: libclang-rt-dev and ninja-build are installed
```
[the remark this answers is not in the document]
Gustavo-Jodar: I put this line while wating for the update for the ```ghcr.io/ledgerhq/ledger-app-builder/ledger-app-dev-tools:latest``` image. I also put a 
```
apt update
apt install -y libclang-rt-dev ninja-build
```
in the local_run.sh script for the same reason. 
Should I remove both of them?

fuzzing/mock/custom/custom_syscalls.c
Gustavo-Jodar: Good call, I will use ```cx_curve_t cv __attribute__((unused))``` then,  this function is in the custom_syscalls.c since they check if the size of the curve key is 32, so for fuzzing I added it. 
Also, the same problem will occur with  the generated_syscalls.c. I can resolve it the same way. Or do you have another idea?

fuzzing/mock/custom/main.c
```diff
@@ -0,0 +1,36 @@
...
+
+#ifdef HAVE_SWAP
+#include "swap.h"
+#endif
+
+#ifdef HAVE_NFC_READER
+#include "os_io_nfc.h"
+#endif  // HAVE_NFC_READER
+
+// TODO: Temporary workaround, at some point all status words should be defined by the SDK and
```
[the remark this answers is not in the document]
Gustavo-Jodar: No, I will remove it, it way copied from /lib_standard_app/io.c. and main.c, forgot to remove them mb.

fuzzing/local_run.sh
Gustavo-Jodar: @bboilot-ledger is it?

fuzzing/libs/glyphs.cmake
Gustavo-Jodar: Nice, will be added.

bboilot-ledger:
FYI, if you want to test the framework with an app, you can have a look at the boilerplate PR https://github.com/LedgerHQ/app-boilerplate/pull/163

fuzzing/mock/custom/mocks.c
```diff
@@ -0,0 +1,33 @@
...
+        }
+        return i;
+    }
+
+    for (i = 0; i < size - 1 && src[i]; i++) {
+        dst[i] = src[i];
+    }
+
+    dst[i] = '\0';  // NUL-terminate
+    while (src[i]) {
```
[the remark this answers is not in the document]
Gustavo-Jodar: To mock that function accordingly with the documentation of [strcpy](https://linux.die.net/man/3/strlcpy) , it says in the "return values" that: 
The strlcpy() and strlcat() functions return the total length of the string they tried to create. For strlcpy() that means the length of src. For strlcat() that means the initial length of dst plus the length of src. While this may seem somewhat confusing, it was done to make truncation detection simple.

I just wanted to stick with what used to be done, but I could change that

fuzzing/libs/glyphs.cmake
```diff
@@ -0,0 +1,54 @@
...
+set(GLYPHS_C ${GEN_GLYPHS_DIR}/glyphs.c)
+set(GLYPHS_H ${GEN_GLYPHS_DIR}/glyphs.h)
+
+# Add target to generate glyphs
+set(GEN_GLYPHS_CMD "${BOLOS_SDK}/lib_nbgl/tools/icon2glyph.py")
+
+file(MAKE_DIRECTORY ${GEN_GLYPHS_DIR})
+add_custom_command(
+  OUTPUT ${GLYPHS_C} ${GLYPHS_H}
+  COMMAND ${Python3_EXECUTABLE} ${GEN_GLYPHS_CMD} ${GLYPH_OPT} --glyphcheader
```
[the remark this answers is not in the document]
0pendev: The fuzzing logic does not implement the full logic of glyphs generation anyway (only supporting nbgl+(stax|flex).
It sounds reasonable to keep it that way here. An easy alternative would be to have a statically generated glyph included as a mock if we end up having to add more logic here (which does not need to be the case IMO).

fuzzing/mock/custom/mocks.c
```diff
@@ -0,0 +1,33 @@
...
+size_t strlcpy(char *dst, const char *src, size_t size)
+{
+    size_t i = 0;
+
+    if (size == 0) {
+        // Can't copy anything or null-terminate
+        while (src[i]) {
+            i++;
+        }
+        return i;
```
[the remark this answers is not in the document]
Gustavo-Jodar: I think it is the same answer as the other comment about [strcpy](https://linux.die.net/man/3/strlcpy) , could I resolve this comment ?

fuzzing/macros/macros.cmake
```diff
@@ -17,16 +17,32 @@ if(NOT "${BOLOS_SDK}/fuzzing" STREQUAL ${CMAKE_SOURCE_DIR})
...
-    file(STRINGS "${CMAKE_SOURCE_DIR}/macros/macros-flex.txt" MACRO_LIST)
+  if(${TARGET_DEVICE} STREQUAL "stax" OR ${TARGET_DEVICE} STREQUAL "flex")
+    # Run `make list-defines` and capture output
+    execute_process(
+      COMMAND sh -c "BOLOS_SDK=\${STAX_SDK} && TARGET=${TARGET_DEVICE} && CC=clang make list-defines"
+      WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}/macros"
+      OUTPUT_VARIABLE MACRO_OUTPUT
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+      ERROR_QUIET
+    )
```
0pendev, line 28: Let's try to get as far as possible from platform specific commands.
Cmake has a few commands/functinos for such use cases.
Here we could try something like cmake env

```suggestion
execute_process(
  COMMAND ${CMAKE_COMMAND} -E env
    BOLOS_SDK=${STAX_SDK}
    TARGET=${TARGET_DEVICE}
    CC=clang
    make list-defines
  WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}/macros"
  OUTPUT_VARIABLE MACRO_OUTPUT
  OUTPUT_STRIP_TRAILING_WHITESPACE
  ERROR_QUIET
)
```

0pendev (changes requested):

fuzzing/local_run.sh
```diff
@@ -0,0 +1,323 @@
...
+REGENERATE_MACROS=0
+RUN_CRASH=0
+SANITIZER="address"
+BOLOS_SDK=""
+FUZZING_PATH="$(pwd)"
+NUM_CPUS=1
+FUZZER=""
+FUZZERNAME=""
+if [ ! -f "$FUZZING_PATH/local_run.sh" ]; then
+    APP_BUILD_PATH="$(cd /app && ledger-manifest -ob ledger_app.toml)"
```
0pendev, line 21: It's fine to have an app_build_path variable but why do we need the hardcoded /app in the rest of the script?
Shouldn't we compute an absolute path here and never use /app again?

wdyt?

fuzzing/README.md
```diff
@@ -2,32 +2,123 @@
...
+./local_run.sh --build=1 --TARGET_DEVICE=stax --fuzzer=build/fuzz_bip32 --j=4 --run-fuzzer=1 --compute-coverage=1
 ```
 
-### Compilation
+### About local_run.sh
+
+| Parameter              | Type                | Description                                                          |
+| :--------------------- | :------------------ | :------------------------------------------------------------------- |
+| `--TARGET_DEVICE`      | `flex or stax`      | **Optional**. Whether it is a flex or stax device (default: flex)    |
+| `--BOLOS_SDK`          | `PATH TO BOLOS SDK` | **Required**. Path to the BOLOS SDK                                  |
```
cedelavergne-ledger, line 57: Most probably the associated variable could be retrieved directly from the `TARGET_DEVICE`, because the SDKs are all under `/opt` in the docker.
The only thing is to convert `nanosp` or `nanos+` into nanos2, and `apex_p` into `apex`

fuzzing/local_run.sh
cedelavergne-ledger: Sounds good for me :thinking:

fuzzing/local_run.sh
```diff
@@ -0,0 +1,323 @@
...
+            make clean BOLOS_SDK="$STAX_SDK"
+            bear --output "$FUZZING_PATH/macros/generated/used_macros.json" -- make -j"$NUM_CPUS" BOLOS_SDK="$STAX_SDK"
+            ;;
+        *)
+            echo -e "${RED}Unsupported TARGET_DEVICE: $TARGET_DEVICE${NC}"
+            exit 1
+            ;;
+    esac
+    cd "$FUZZING_PATH/macros/" || exit 1
+    python3 "$BOLOS_SDK/fuzzing/macros/extract_macros.py" \
```
cedelavergne-ledger, line 103: Seems `BOLOS_SDK`, here, could be different from the `case`/`esac` above, depending on the selected device?

fuzzing/local_run.sh
```diff
@@ -0,0 +1,323 @@
...
+            show_help
+            ;;
+        *)
+            echo -e "${RED}Unknown argument: $arg${NC}"
+            show_help
+            ;;
+    esac
+done
+
+#===============================================================================
```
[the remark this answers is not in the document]
0pendev: Could be usefull to deduce TARGET_DEVICE from `BOLOS_SDK` so that one can still use it's own version of the ledger-secure-sdk for fuzzing. 

@cedelavergne-ledger what do you think about setting `BOLOS_SDK`, searching for .target and if `.target` does not exists either throw an error message or check for a `--TARGET_DEVICE` ?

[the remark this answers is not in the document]
cedelavergne-ledger: I understand.
Yes, your proposal makes sense !

Just another subsidiary question about that: why did you define the name `TARGET_DEVICE`, whereas the SDK uses already `TARGET` (for `stax`, `flex`...) and also `TARGET_NAME` (for `TARGET_STAX`, `TARGET_FLEX`...)?

fuzzing/CMakeLists.txt
```diff
@@ -21,9 +21,16 @@ if(NOT CMAKE_C_COMPILER_ID MATCHES "Clang")
   message(FATAL_ERROR "Fuzzer needs to be built with Clang")
 endif()
 
-# default fuzz device target
-if(NOT TARGET_DEVICE)
-  set(TARGET_DEVICE "flex")
+# fuzz device target
```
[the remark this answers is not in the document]
cedelavergne-ledger: I think it is better, avoiding the user to select a SDK, with an incompatible device (for example, when we change the API_LEVEL)

But according to another question, one may want to use its own SDK, and still provide a TARGET (or TARGET_DEVICE) to switch for example betwen flex and stax, without changing the SDK

Makefile.rules
```diff
@@ -123,3 +123,10 @@ ifneq ($(wildcard $(APP_MANIFEST_FILE)),)
 endif
 
 include $(BOLOS_SDK)/Makefile.rules_generic
+
+list-defines:
```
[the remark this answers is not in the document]
cedelavergne-ledger: Apps should be aligned with Standard App framework. This is the rule to be able to use the fuzzer

.clusterfuzzlite/build.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -e
+
+cat <<EOF >.target
```
cedelavergne-ledger, line 3: I think you you check 1st if the file exist, to not overwrite it!
Besides, I am wondering if it is safer to rather use
`export TARGET=flex`
in order to keep the SDK without extra modification...

cedelavergne-ledger (approved):
Good job!

PR #976, opened 2025-05-22
Fix issue when navigating without settings page

cedelavergne-ledger:


iartemov-ledger (approved):
It seems to me fine and it works on my side with only "App info".

PR #982, opened 2025-05-26
Reverse App-Home icon, and apply rounded corners

cedelavergne-ledger:
Add script to convert the App Icon for home screen (white background), and apply rounded corners.

The principle consists in declaring the destination file in the App Makefile:
`ICON_HOME_NANO = glyphs/home_icon_14px.gif`

If the variable `ICON_HOME_NANO` is found, the corresponding glyph file is generated, using the App icon as input (`ICONNAME` in SDK, initialized from `ICON_NANOX` or `ICON_NANOSP` in App Makefile).

PR #983, opened 2025-05-26
Reverse App-Home icon, and apply rounded corners - API_LEVEL_22

cedelavergne-ledger:
(cherry picked from https://github.com/LedgerHQ/ledger-secure-sdk/pull/982)

### DIFF
--- a/Makefile.glyphs
+++ b/Makefile.glyphs
@@ -75,7 +75,15 @@ $(GLYPH_DESTH): $(GLYPH_FILES) $(ICON_SCRIPT)
 	$(L)echo "[GLYPH] Compiling..."
 	$(L)-mkdir -p $(GLYPH_SRC_DIR)
 ifneq ($(USE_NBGL),0)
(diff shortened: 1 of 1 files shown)


iartemov-ledger (approved):
We discussed it a lot. 

It is not a perfect solution (none is):
* as we mix up the inputs and outputs in  Makefile of app - which is a bit confusing
* app generates outputs outside of `build` folder


Another solution was proposed:
* not to declare `ICON_HOME_NANO` in app's `Makefile`
* but in `Makefile` of SDK  to generate an image inverted to ICONNAME  in `build/target` sub-folder
* after: 
     * either it would generate  `C_<internal_name>` in `glyphs.c` and in `glyphs.h`  a macro  ICON_APP_HOME (e.g.) pointing to this variable and then proposed to developer to be used in `nbgl_useCaseHomeAndSettings()`
     * or it would generated `C_<initial_name>_inv` to be used in a hidden way by SDK if `appIcon` parameter passed to `nbgl_useCaseHomeAndSettings()` is NULL
     
     This solution was refused as hard-coded names would be used. 
     
     So approving !

PR #1068, opened 2025-07-31
Several fixes for recent SDK changes

iartemov-ledger:
## Description
- [x] a couple of  Clang static analyzer fixes
- [x]  more serious issus when `const uint8_t status[2] = {0x69, 0x85}` was not taken into account when located under the if condition. See the assembly comparison screenshot  - on the left before the fix and on the right after the fix We clearly see the 0x6985 assignment).  This impacts SK and its tests.

Screenshot from 2025-08-01 10-15-58

## Changes include

- [ ] Tests
- [ ] Documentation

## Breaking changes

## Additional comments


jarevalo-ledger (approved):
LGTM

PR #1036, opened 2025-07-08
Regain control over the SDK libraries

iartemov-ledger:
:warning: **Current goal is to get first feedback, the discussions and the tests are still on-going**

Based on @apaillier-ledger's work here: https://github.com/LedgerHQ/ledger-secure-sdk/pull/924.

Corresponds to https://ledgerhq.atlassian.net/browse/B2CA-1995.
See https://github.com/LedgerHQ/ledger-secure-sdk/blob/feat/regain_control_over_libs/arch/libraries.md as documentation.
Also internal documentation:
* https://ledgerhq.atlassian.net/wiki/spaces/FW/pages/5775458422/WIP+Regain+control+over+the+libraries+used+in+the+SDK
* https://ledgerhq.atlassian.net/wiki/spaces/FW/pages/6009389224/Choice+for+libc+and+libm

- [x] `libclang_rt.builtins` replaces `libgcc` for Nano X, Stax and Flex targets
- [x] it is built using `.github/workflows/build_clangrt_builtins.yml` workflow and pushed through https://github.com/LedgerHQ/ledger-secure-sdk/pull/1035
- [x] `libc` and `libm` removed from the SDK and from now on the ones from `picolibc` Debian package are used 
- [ ]  to test using `ledger-app-tester`
    - [x] see e.g. https://github.com/LedgerHQ/ledger-app-tester/actions/runs/16142164572, where a special branch  based on SDK v22.14.0 is used with cherry-picked commits from the present branch: https://github.com/LedgerHQ/ledger-secure-sdk/tree/feat/22_regain_control_over_libs
    - [x] size impact 
    - [ ] performance impact
- [ ]  to test on physical device 
- [ ] several apps builds are broken (by the `SYSROOT` change, most probably because of incorrect header inclusion) 
    - [ ] `app-plugin-rarible`
    - [ ] `app-tron` (fix example: https://github.com/LedgerHQ/app-tron/pull/88)
    - [ ] `app-mimblewimblecoin`
    - [ ] `app-plugin-okx`
    - [ ] `app-acre`
    - [ ] `app-zcash-new`
- [ ] must be merged together with corresponding Rust SDK PR as it breaks the latter: https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/269

## Description

Please provide a detailed description of what was done in this PR.
(And mention any links to an issue [docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue))

## Changes include

[...]

PR #1076, opened 2025-08-05
Regain control over the SDK libraries [new variant using rebuilt newlib from Debian sources)

iartemov-ledger:
# Description

Replacement for https://github.com/LedgerHQ/ledger-secure-sdk/pull/1036
Corresponds to https://ledgerhq.atlassian.net/browse/B2CA-1995.

## DONE
- [x] `libclang_rt.builtins` replaces `libgcc` for Nano X, Stax and Flex targets, built using `build_clangrt_builtins.sh` script
- [x] `libc` and `libm` have been updated from the ones built using `build_newlib.sh` script (newlib 3.3.0 from Debian 12 currently used for `app-builder`)
- [x] the both are built using .github/workflows/build_clangrt_builtins.yml workflow (https://github.com/LedgerHQ/ledger-secure-sdk/actions/runs/17647456726 job and pushed through https://github.com/LedgerHQ/ledger-secure-sdk/pull/1186 - it matches to the libs built from `feat/regain_control_over_libs_2`)
- [x] Rust part is not be impacted - tested locally with `app-boilerplate-rust` and `app-starknet` apps. This is expected as Rust build does not seem to use `libgcc.a` and `libm.a` - only `libc.a`. So some improvements to be done later in the Rust SDK as follow https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/269/
- [x] remove `libgcc` from the SDK (still "used" by NanoS)?  - let's do it in an independent PR cleaning up all occurrences! 

##  Tests
- [x] using API_LEVEL_24 derived branch and https://github.com/LedgerHQ/ledger-app-tester/actions/runs/17665725023 tests - see per app size comparison: https://docs.google.com/spreadsheets/d/10lRJbRVR3c8e0jBQqgb4xu2yBWZQiNXEoVCkfQiI1kw/edit?pli=1&gid=1027262762#gid=1027262762
- [x]  functions used in `app-ethereum`  - see [[NEW] lib.c.newlib.rebuilt.armv-v8m] column in https://docs.google.com/spreadsheets/d/1qI_2yP-VMc6ze2DHbfCj5uHf3eDwh5NnysEewkuJuBw/edit?gid=1206824572#gid=1206824572
- [x]  performance impact  - not visible with manual local tets. Tested also by comparing test execution time: https://docs.google.com/spreadsheets/d/10lRJbRVR3c8e0jBQqgb4xu2yBWZQiNXEoVCkfQiI1kw/edit?pli=1&gid=557969334#gid=557969334 
- [x]  to test on physical device (`NanoSP`, `Flex`, `app-bitcoin-new`, `app-boilerplate-rust`, `app-starknet`)

## Few other remarks
 - the libs in question are not used for he Ledge rOS build
 - once we switch to Debian 13 in `app-builder` we'll need to rebuild the libs and benefit  among other from the newest newlib version 4.5.0

## Changes include

- [ ] Tests
- [ ] Documentation

## Breaking changes

## Additional comments

[x] TARGET_API_LEVEL: API_LEVEL_24
[x] TARGET_API_LEVEL: API_LEVEL_25


tools/build_newlib.sh
```diff
@@ -0,0 +1,97 @@
...
+cd "${WORKDIR}"
+
+# Installing newlib build dependencies
+apt update
+apt install -y --no-install-recommends texinfo
+
+# Get the latest fixed version
+git clone --depth=1 --single-branch --branch newlib-4.5.0 https://sourceware.org/git/newlib-cygwin.git "${SRC}"
+
+# Apply patch to fix "undefined symbol: __aeabi_unwind_cpp_pr0 referenced by libc_a-setjmp.o" issue
```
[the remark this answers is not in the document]
iartemov-ledger: Thank you @0pendev .
The advantage of your solution is that is the disadvantage of mine - we do not need to patch the code that might be updated so the patch does not apply any more. 
Though it remains heavy (extract -> remove -> re-compress). 

In ideal world it would be a newlib compilation flag switching off the unwind/.ARM.exidx section , but I did not find it. 
I will try and test your solution.

tools/build_newlib.sh
```diff
@@ -0,0 +1,97 @@
...
+cp "${SCRIPT_DIR}/${PATCH_NAME}" "${SRC}"
+cd "${SRC}"
+git apply "${PATCH_NAME}"
+cd ..
+
+# Build configuration
+mkdir -p "${BUILD}" "${INSTALL}"
+cd "${BUILD}"
+
+CFLAGS_FOR_TARGET="-ffunction-sections -fdata-sections -fshort-wchar -DPREFER_SIZE_OVER_SPEED -mcpu=${TARGET_CPU} -mthumb -mlittle-endian" \
```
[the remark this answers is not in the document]
iartemov-ledger: > the generated library is bigger overall (70Ko bigger).
@ :sweat_smile:
Plus need to bring the newer clang.

Honestly I do not know if it it worth it.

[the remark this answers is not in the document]
0pendev: Agreed :raised_hands: Could be something we look into next time we update the Debian image of the ledger-app-builder.

.github/workflows/build_clangrt_builtins.yml
```diff
@@ -0,0 +1,99 @@
...
+        type: string
+        required: false
+        default: 'master'
+      create_pr:
+        type: boolean
+        required: false
+        default: false
+
+  # TODO: remove this next line before merging!
+  push:
```
0pendev, line 18: @bboilot-ledger should we remove this?

tools/build_newlib.sh
```diff
@@ -0,0 +1,91 @@
...
+apt install -y --no-install-recommends texinfo
+
+# Get the latest fixed version
+git clone --depth=1 --single-branch --branch newlib-4.5.0 https://sourceware.org/git/newlib-cygwin.git "${SRC}"
+
+# Build configuration
+mkdir -p "${BUILD}" "${INSTALL}"
+cd "${BUILD}"
+
+CFLAGS_FOR_TARGET="-ffunction-sections -fdata-sections -fshort-wchar -DPREFER_SIZE_OVER_SPEED -mcpu=${TARGET_CPU} -mthumb -mlittle-endian" \
```
bboilot-ledger, line 53: Should we use `-fPIC` here?

.github/workflows/build_clangrt_builtins.yml
iartemov-ledger: I'm not 100% sure, but I think we should not as:
* with the static library we build finally an executable  (an app) and we do not use `-fPIC` for the app build
* we do not target any shared libraries 
* `-fPIC`-generated code is bit less optimized and everything is working today without it ;)

tools/build_newlib.sh
```diff
@@ -0,0 +1,91 @@
...
+BUILD="${SRC}/arm_none_eabi_build"
+INSTALL="${SRC}/arm_none_eabi_install"
+cd "${WORKDIR}"
+
+# Installing newlib build dependencies
+apt update
+apt install -y --no-install-recommends texinfo
+
+# Get the latest fixed version
+git clone --depth=1 --single-branch --branch newlib-4.5.0 https://sourceware.org/git/newlib-cygwin.git "${SRC}"
```
apaillier-ledger, line 47: I would prefer if the newlib sources were downloaded with `apt source` like how it is done for LLVM in the other script.
That way it would have the same exact version as the one already installed.
Because like this, an app build would use the newlib 4.5.0 static libraries but the 3.3.0 header files.
If newlib 4.5.0 is somehow required, maybe we should wait for the migration of the app-builder container to Debian 13.

tools/build_newlib.sh
iartemov-ledger: This topics may need some deeper investigation, but with a lower priority IMHO.
As today we do not observe these quirks.
Plus the initial .a  were built without `fPIC` I believe (by observation).

.github/workflows/build_clangrt_builtins.yml
iartemov-ledger: removed

apaillier-ledger (approved):

PR #1149, opened 2025-09-02
Fix whole screen issue when building screenshots

dmorais-ledger:
## Description

The screen resolution on APEX is much smaller than on Flex/Stax, which causes problems when displaying text across the entire screen.

When such a case occur, LOG_FATAL function is called, causing the build to fail.
When we build screenshots, we don't want the build to fail, we want an error message to be displayed instead, making it possible for translators to make the text shorter.

This PR fix that issue.

## Changes include

### DIFF
--- a/lib_nbgl/src/nbgl_obj.c
+++ b/lib_nbgl/src/nbgl_obj.c
@@ -353,21 +353,38 @@ static void compute_position(nbgl_obj_t *obj, nbgl_obj_t *prevObj)
 
     obj->area.x0 = parent->obj.area.x0 + obj->rel_x0;
     obj->area.y0 = parent->obj.area.y0 + obj->rel_y0;
-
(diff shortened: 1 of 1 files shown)


lib_nbgl/src/nbgl_obj.c
```diff
@@ -355,19 +355,37 @@ static void compute_position(nbgl_obj_t *obj, nbgl_obj_t *prevObj)
     obj->area.y0 = parent->obj.area.y0 + obj->rel_y0;
 
     if ((obj->area.x0 + obj->area.width) > SCREEN_WIDTH) {
```
nroggeman-ledger, line 357: You can change to `if ((obj->area.x0 + obj->area.width) > (parent->obj.area.x0 + parent->obj.area.width)) `

nroggeman-ledger (changes requested):

nroggeman-ledger (approved):

PR #1215, opened 2025-09-18
Replace 'x of y' by 'x / y' on multi pages navigation

dmorais-ledger:
## Description

Replace the text "x of y" by "x / y" when navigating multiple pages.
Doing that will make this string usable on all languages.
The result is what is displayed on next picture

image


tdejoigny-ledger:
@dmorais-ledger I would say : let's ask first to product team :)
cc: @nroggeman-ledger

PR #1241, opened 2025-09-26
Avoid to crash when parsing strings with Unicode in Apps

nroggeman-ledger:
## Description

The goal of this PR is to "support" Unicode chars in Apps, by considering them as "empty", instead of crashing when trying to get them from Language pack when encountered.
This fix will be in shared.elf so needs to be integrated in OS to be used.

## Changes include

- [ ] Tests
- [ ] Documentation

## Breaking changes

## Additional comments

[x] TARGET_API_LEVEL: API_LEVEL_24
[x] TARGET_API_LEVEL: API_LEVEL_25

Remember:

### DIFF
--- a/lib_nbgl/src/nbgl_fonts.c
+++ b/lib_nbgl/src/nbgl_fonts.c
@@ -13,6 +13,7 @@
 #include "nbgl_fonts.h"
 #include "os_helpers.h"
 #include "os_pic.h"
+#include "os_task.h"
 #if defined(HAVE_LANGUAGE_PACK)
 #include "ux_loc.h"
 #endif  // defined(HAVE_LANGUAGE_PACK)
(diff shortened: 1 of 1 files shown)


lib_nbgl/src/nbgl_fonts.c
```diff
@@ -1246,6 +1247,10 @@ void nbgl_textReduceOnNbLines(nbgl_font_id_e fontId,
  */
 nbgl_unicode_ctx_t *nbgl_getUnicodeFont(nbgl_font_id_e fontId)
 {
+    // if in Apps, do not support unicode
+    if (os_sched_current_task() != TASK_BOLOS_UX) {
+        return NULL;
```
iartemov-ledger, line 1252: As this change will only make part of next OS release may be we have time for a more elegant solution.
Would it be possible:
* for a sub-set of selected characters, for example making part of European alphabets (like Â, É, à, é, ï etc.) and with which we've had already issues, to make a dictionary conversion to counterpart selected without accents.  
* and keep skipping others 
?

lib_nbgl/src/nbgl_fonts.c
```diff
@@ -1246,6 +1247,10 @@ void nbgl_textReduceOnNbLines(nbgl_font_id_e fontId,
  */
 nbgl_unicode_ctx_t *nbgl_getUnicodeFont(nbgl_font_id_e fontId)
 {
+    // if in Apps, do not support unicode
```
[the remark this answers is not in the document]
iartemov-ledger: Friday evening mistake - there is a loop each time and the variable is verified din the loop to check if we need to re-charger the Unicode context.

lib_nbgl/src/nbgl_fonts.c
nroggeman-ledger: It's an app choice to do substitutions...

PR #1247, opened 2025-09-30
Avoid to crash when parsing strings with Unicode in Apps

nroggeman-ledger:
## Description

The goal of this PR is to "support" Unicode chars in Apps, by considering them as "empty", instead of crashing when trying to get them from Language pack when encountered.
This fix will be in shared.elf so needs to be integrated in OS to be used.

(Cherry-pick from master)

## Changes include

- [ ] Tests
- [ ] Documentation


iartemov-ledger:
Why this is not a cherry-pick from https://github.com/LedgerHQ/ledger-secure-sdk/pull/1241 ?

PR #1514, opened 2026-04-22
Remove OS-specific header includes

yrichard-ledger:
## Description

To cleanup the build system, remove some includes that are only used by the OS and have no place in the app SDK

## Changes include

- [ ] Tests
- [ ] Documentation

## Breaking changes

## Additional comments

[ ] TARGET_API_LEVEL: API_LEVEL_25
[ ] TARGET_API_LEVEL: API_LEVEL_26

Remember:


include/os_utils.h
```diff
@@ -1,9 +1,12 @@
 #pragma once
 
 #include <stdbool.h>
-#include "arch.h"
+#include <stddef.h>
+#include <stdint.h>
 #include "decorators.h"
 
+#define WIDE
```
yrichard-ledger, line 8: This definition is still used by some applications so I moved it here

jarevalo-ledger (approved):

PR #1566, opened 2026-05-29
some fixes to comply with clang-21

mbrousset-ledger:
## Description

Prepare toolchain update (clang-14 to clang-21)
This PR is needed by related PR in `ledger-app-builder` : https://github.com/LedgerHQ/ledger-app-builder/pull/226 (which CI won't pass until this one gets merged)

## Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)
- [ ] Tests
- [ ] Documentation
- [x] Other (for changes that might not fit in any category)

## Breaking changes

Please complete this section if any breaking changes have been made, otherwise delete it.

## Additional comments

Please post additional comments in this section if you have them, otherwise delete it.

## Auto cherry-pick in API_LEVEL

If requested to port the commits from this PR on a dedicated _API_LEVEL_ branch,
select the targeted one(s), or add new references if not listed:

[x] TARGET_API_LEVEL: API_LEVEL_26

This will only create the PR with cherry-picks, ready to be reviewed and merged.

Remember:

- The merge will ALWAYS be a manual operation.
- It is possible the cherry-picks don't apply correctly, mainly if previous commits have been forgotten.
- In case of failure, there is no other solution than redo the operation manually...

PR #1567, opened 2026-06-01
Use 'unsigned int' instead  of 'uint'

bigspider:
## Description

`uint` is not a standard type.

## Changes include

- [ ] Tests
- [ ] Documentation

Remember:


bigspider:
Same fix is already also in #1566

PR #1573, opened 2026-06-03
Fix unit-tests workflow

apaillier-ledger:
## Description

Following a recent app-builder container update.

## Changes include

- [ ] Tests
- [ ] Documentation

Remember:

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -20,8 +20,10 @@ jobs:
       - name: Clone
         uses: actions/checkout@v6
 
-      - name: Install package libbsd-dev
-        run: apt-get install -y libbsd-dev curl gpg
+      - name: Install curl
+        run: |
+          apt-get update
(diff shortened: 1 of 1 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -21,7 +21,9 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Install package libbsd-dev
-        run: apt-get install -y libbsd-dev curl gpg
+        run: |
+          apt-get update
+          apt-get install -y libbsd-dev curl gpg
```
mbrousset-ledger:
```suggestion
          apt-get install -y libbsd-dev curl
```

`gpg` is already installed in app-builder-lite

apaillier-ledger: Good catch, also removed libbsd-dev :+1:

cedelavergne-ledger (approved):

PR #1578, opened 2026-06-03
add stack protector

mbrousset-ledger:
## Description

- [x] Adding `ENABLE_STACK_PROTECTOR` option. This is an opt-in for apps that will define `ENABLE_STACK_PROTECTOR=1` in their `Makefile`.

PoC from #1128 by @bboilot-ledger 

## Changes include

- [ ] Tests
- [ ] Documentation

## Breaking changes

## Additional comments

Remember:

### DIFF
--- a/target/stax/script.ld
+++ b/target/stax/script.ld
@@ -48,7 +48,9 @@ SECTIONS
     _text = .;
     _nvram_start = .;
 
-    /* ensure main is always @ 0xC0D00000 */
+    /* ensure __stack_chk_init is always @ 0xC0D00000 */
+    KEEP(*(.boot.ssp_init))
+    /* ensure main directly follows __stack_chk_init */
     *(.boot*)
 
     /* place the other code and rodata defined BUT nvram variables that are displaced in a r/w area */
@@ -108,6 +110,9 @@ SECTIONS
     /**
      * Place RAM uninitialized variables
      */
+    __stack_chk_guard = .;
+    PROVIDE(__stack_chk_guard = .);
+    . += 4;
     _bss = .;
     *(.bss*)
     _ebss = .;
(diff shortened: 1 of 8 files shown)


bboilot-ledger (approved):

target/stax/script.ld
```diff
@@ -108,6 +110,9 @@ SECTIONS
     /**
      * Place RAM uninitialized variables
      */
+    __stack_chk_guard = .;
```
iartemov-ledger, line 113: If something moves in the solution (`__stack_chk_guard` is not the 1st .bss variable or `__wrap___stack_chk_init()` is not 1st function in `.text`) it does not work any more. 
From this PoV should not we have a non-regression test (not sure which form, in a test app?) that:
- checks that the stack protection, as it is from the SDK, works
- that `__stack_chk_guard` is randomized 
?

target/stax/script.ld
```diff
@@ -130,6 +135,12 @@ SECTIONS
 
   ASSERT( (_estack - _stack) >= STACK_MIN_SIZE, "stack section too small" )
 
+  /* The .init_array is initialized with functions with the constructor
+   * attribute. Discard this section since there's no loader. */
```
iartemov-ledger, line 139: This change in all linker scripts is not related the stack protection, is it?

iartemov-ledger (commented):
A couple of questions stp.

target/stax/script.ld
bboilot-ledger: If I remember correctly, adding the stack protector flag set this array with a non empty value so the canary can be initialized before the main function using the crt constructor mechanism. But since we don't support it (no loader, entrypoint enforced etc), we must remove it otherwise the app will crash.

iartemov-ledger (approved):
Approving the remark https://github.com/LedgerHQ/ledger-secure-sdk/pull/1578/files#r3381156890.

As proposed by @mbrousset-ledger  to address above as a separate task we may explorer a link-time or an assert-in-linker-script tracks.

PR #1588, opened 2026-06-08
 fix TOCTOU RC: reject incoming APDU if there is  already a pending one

iartemov-ledger:
## Description

**TOCTOU consent latch to the legacy IO loop.**

When the app accepts a command that defers its reply (e.g. a transaction review), it sets a reply_pending latch (`os_io_set_reply_pending(true)`); the reply path clears it (in `io_legacy_apdu_tx` and `u2f_message_reply`).
While the latch is set, any new incoming APDU is rejected with `SWO_COMMAND_NOT_ACCEPTED (0x6901)` instead of being processed.

U2F/CTAP framing is deliberately exempt from the reject gate (`CANCEL/RAW/CBOR`), with CBOR still arming/clearing the latch like a normal command — so FIDO traffic isn't disturbed (`CTAPHID` does its own `CHANNEL_BUSY` arbitration above this layer).

### The fix, in one paragraph

A static `bool io_reply_pending` (latch) in `io_legacy/src/os_io_legacy.c`:
- **armed** when a command is accepted in `io_legacy_apdu_rx` (excludes CANCEL/RAW);
- **reject gate**: while armed, a new command is refused with `SWO_COMMAND_NOT_ACCEPTED`
  (`0x6901`) (exempts CANCEL/RAW/CBOR);
- **cleared** on the app's reply — `io_legacy_apdu_tx` (sync + async/deferred) and
  `u2f_message_reply` (U2F/CTAP, unconditional).

This hooks the shared `io_legacy_apdu_rx` chokepoint, so it covers every app and every
deferred-UI window without per-app changes.

### Findings rebuttal

| # | Finding (their claim) | Severity claimed | Why we reject it |
|---|---|---|---|
| **H1** | CBOR frame passes the reject gate, is answered by `u2f_message_reply`, which clears the latch unconditionally → TOCTOU window reopens behind a pending review | HIGH / "CRITICAL if FIDO2" | **The exploit needs one app that is both a signer (deferred APDU review) and a FIDO authenticator (processes CBOR). No shipping app is both** — signing apps (Monero, Bitcoin, Eth, Solana…) don't process CBOR; the FIDO app has no deferred APDU review. So the unconditional clear in `u2f_message_reply` can never erase a *genuine* pending review. Documented in-code with a "make armer-aware if such an app ever ships" pointer. The proposed enum adds fragility to guard a combination that cannot occur. |
| **H1-bis** (CBOR→CBOR case) | 1st CBOR arms latch + starts UI; 2nd CBOR is exempt from the reject gate → overwrites the shared `G_io_apdu_buffer` behind the UI | (implied by H1) | **True at the latch level — the latch does *not* refuse the 2nd CBOR — but the 2nd CBOR never reaches it.** The CTAPHID transport serializes one transaction per handle: while the 1st is in process, any non-CANCEL frame gets `CTAP1_ERR_CHANNEL_BUSY` at `lib_u2f/src/u2f_transport.c:81-91` and is never reassembled into a delivered 0x24 packet. So no second `memmove` into the shared buffer ever happens — no overwrite, no TOCTOU. |
| **H2** | A command that arms but never replies leaves the latch stuck `true` → all later APDUs rejected `0x6901` until power-cycle | MEDIUM | **Self-healing, not a lock.** Every normal reply path clears the latch (`io_legacy_apdu_tx` for sync *and* async/deferred replies, `u2f_message_reply` for U2F) — the review's own table admits this. The "fire-and-forget / `SWO_NO_RESPONSE` without TX" path doesn't exist in the legacy dispatch model: an accepted command always ends in an SW, and any non-OK SW exits to dashboard (state reset). A wedge would be a self-DoS recoverable by disconnect, not a security defect. |
| **W2** | The two exemption lists are asymmetric — arm excludes only {CANCEL, RAW}, reject excludes {CANCEL, RAW, CBOR} | WARNING (claimed root cause of H1) | **Intentional division of labor, not a bug.** CANCEL/RAW are transport-only frames carrying no command → never arm, never gated. CBOR *is* a command (it replies like one) so it arms, but it's left exempt from the reject gate **on purpose**: CBOR↔CBOR concurrency is already serialized one layer down by CTAPHID (`CHANNEL_BUSY`, `u2f_transport.c:81`), which also preserves the legitimate CANCEL/keepalive flow. Having the latch *also* reject CBOR would be redundant and would force it to re-implement the CANCEL exception to avoid breaking FIDO. The asymmetry is correct per frame type. |
| **W3** | No negative test for CBOR interleaving during a pending review | WARNING | **Nothing reachable to test.** The interleave only exists if H1/H1-bis are reachable, and they aren't (no signer+FIDO app; CBOR→CBOR serialized by the transport). A test would have to fabricate a non-existent app config. The real surface — APDU-interleave-during-review — is covered by the app-monero V-030 ragger tests. |
| **W1** | The reject response reuses the incoming packet type | WARNING (observation) | **Correct by design.** The `0x6901` rejection must go back to the host on the transport that sent the offending frame (`os_io_legacy.c:410`). Standard SDK reply behavior, not a flaw. |

### Test run with the same fix applied to API_LEVEL_26-derived branch (`reject_if_pending_26`): 
- https://github.com/LedgerHQ/ledger-app-tester/actions/runs/27147141553  :green_circle: 
- https://github.com/LedgerHQ/ledger-app-tester/actions/runs/27222643702 :green_circle: 
- https://github.com/LedgerHQ/ledger-app-tester/actions/runs/27267499532 :green_circle: 

## Changes include

- [ ] Tests
- [ ] Documentation

## Breaking changes

It may break:
- dishonest that sends a 2nd APDU while user is on conf
- some "honest" uses cases if this unusual behavior is really anticipated (to be found) 

## Additional comments

Remember:

### DIFF
--- a/io_legacy/src/os_io_legacy.c
+++ b/io_legacy/src/os_io_legacy.c
@@ -80,6 +80,8 @@ extern unsigned int app_stack_canary;
 // Variable containing the type of the APDU response to send back.
 static apdu_type_t io_os_legacy_apdu_type;
 static uint8_t     need_to_start_io;
+// TOCTOU latch
+static bool io_reply_pending;
 
 /* Private functions ---------------------------------------------------------*/
 static io_apdu_media_t get_media_from_apdu_type(apdu_type_t apdu_type)
@@ -392,6 +394,22 @@ int io_legacy_apdu_rx(uint8_t handle_ux_events)
             case OS_IO_PACKET_TYPE_USB_U2F_HID_RAW:
             case OS_IO_PACKET_TYPE_BLE_APDU:
             case OS_IO_PACKET_TYPE_NFC_APDU:
+                // TOCTOU: reject a new command while a previous one still awaits its reply.
+                // CANCEL/RAW are pure U2F transport frames and never touch the latch; CBOR is
+                // exempt from rejection here but still arms/clears it like a normal command.
+                if ((G_io_rx_buffer[0] != OS_IO_PACKET_TYPE_USB_U2F_HID_CANCEL)
+                    && (G_io_rx_buffer[0] != OS_IO_PACKET_TYPE_USB_U2F_HID_RAW)
+                    && (G_io_rx_buffer[0] != OS_IO_PACKET_TYPE_USB_U2F_HID_CBOR)
+                    && os_io_reply_pending()) {
+                    bolos_err_t err   = SWO_COMMAND_NOT_ACCEPTED;
+                    G_io_tx_buffer[0] = err >> 8;
+                    G_io_tx_buffer[1] = err;
+                    status            = os_io_tx_cmd(G_io_rx_buffer[0], G_io_tx_buffer, 2, 0);
+                    if (status > 0) {
+                        status = 0;
+                    }
+                    break;
+                }
                 io_os_legacy_apdu_type = G_io_rx_buffer[0];
                 if (os_perso_is_pin_set() == BOLOS_TRUE
                     && os_global_pin_is_validated() != BOLOS_TRUE) {
@@ -435,6 +453,12 @@ int io_legacy_apdu_rx(uint8_t handle_ux_events)
                 }
 #endif  // HAVE_BOLOS_NO_DEFAULT_APDU
                 else {
+                    // Arm the TOCTOU latch. CANCEL/RAW are transport-only frames with no app
+                    // reply to clear it, so they don't arm; CBOR does (it replies like a command).
+                    if ((G_io_rx_buffer[0] != OS_IO_PACKET_TYPE_USB_U2F_HID_CANCEL)
+                        && (G_io_rx_buffer[0] != OS_IO_PACKET_TYPE_USB_U2F_HID_RAW)) {
+                        os_io_set_reply_pending(true);
+                    }
                     G_io_app.apdu_media = get_media_from_apdu_type(io_os_legacy_apdu_type);
                     status -= 1;
                     memmove(G_io_apdu_buffer, &G_io_rx_buffer[1], status);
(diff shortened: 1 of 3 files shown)


io/include/os_io.h
```diff
@@ -167,6 +167,15 @@ SYSCALL int os_io_seph_se_rx_event(unsigned char *buffer PLENGTH(length),
 
 unsigned int os_io_handle_ux_event_reject_apdu(void);
 
+// CWE-367 reply-pending latch. Armed when an APDU command is accepted; cleared
```
tdejoigny-ledger: let's remove this kind of comments

iartemov-ledger: I've replaced the CWE  by the well-known term TOCTOU, which IMHO is appropriate here since this is what we are protecting against.

lib_u2f_legacy/src/u2f_legacy.c
```diff
@@ -52,6 +52,9 @@ void u2f_message_set_autoreply_wait_user_presence(u2f_service_t *service, bool e
 void u2f_message_reply(u2f_service_t *service, uint8_t cmd, uint8_t *buffer, uint16_t len)
 {
     (void) service;
+    // TOCTOU: the CTAP2/U2F reply bypasses io_legacy_apdu_tx, so release the
+    // latch here too.
+    os_io_set_reply_pending(false);
     if (cmd == U2F_COMMAND_HID_CBOR) {
```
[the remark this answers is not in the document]
iartemov-ledger: Fixed.

io_legacy/src/os_io_legacy.c
```diff
@@ -392,6 +392,20 @@ int io_legacy_apdu_rx(uint8_t handle_ux_events)
...
+                    && os_io_reply_pending()) {
+                    bolos_err_t err   = SWO_COMMAND_NOT_ACCEPTED;
+                    G_io_tx_buffer[0] = err >> 8;
+                    G_io_tx_buffer[1] = err;
+                    status            = os_io_tx_cmd(G_io_rx_buffer[0], G_io_tx_buffer, 2, 0);
+                    if (status > 0) {
+                        status = 0;
+                    }
+                    break;
+                }
```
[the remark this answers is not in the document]
iartemov-ledger: Fixed.

io/include/os_io.h
sportron-ledger, line 1: As it 'legacy" related, why don't put this in os_io_legacy.h?

io/src/os_io.c
sportron-ledger, line 1: As it 'legacy" related, why don't put this in os_io_legacy.c?

io/include/os_io.h
iartemov-ledger: Good point! Moved to `os_io_legacy.{c,h}`.

io/src/os_io.c
iartemov-ledger: Good point! Moved to `os_io_legacy.{c,h}`.

bboilot-ledger (approved):
We might want to notify the DMK/LW team in case they encounter this new status word.

PR #1606, opened 2026-06-17
Address book: replace SDK implementation by syscalls

srasoamiaramanana-ledger:
## Description

This PR adds two syscalls for the Address Book feature.

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/app_features/address_book/src/address_book_crypto.c
+++ b/app_features/address_book/src/address_book_crypto.c
@@ -18,21 +18,12 @@
  * @file address_book_crypto.c
  * @brief HMAC Proof of Registration for Address Book contacts
  *
- * Provides HMAC-SHA256 based proofs that bind a registered contact to this
- * specific device. Two independent key derivations are used, one per feature:
+ * All HMAC key derivation and computation are delegated to two OS syscalls:
+ *   - sys_address_book_hmac()        — compute HMAC
+ *   - sys_address_book_hmac_verify() — verify HMAC (constant-time)
  *
- *  - Identity KDF:       SHA256("AddressBook-Identity"      || privkey.d)
- *  - Ledger Account KDF: SHA256("AddressBook-LedgerAccount" || privkey.d)
- *
- * HMAC messages:
- *  - Group KDF:          SHA256("AddressBook-Group"         || privkey.d)
- *  - Identity HMAC_PROOF: gid(32) | name_len(1) | name
- *  - Identity HMAC_REST: gid(32) | scope_len(1) | scope | id_len(1) | identifier |
- *                        family(1) [| chain_id(8) for FAMILY_ETHEREUM]
- *  - Ledger Account:     name_len(1) | name | family(1) [| chain_id(8) for FAMILY_ETHEREUM]
- *
- * The proof is stored by the host and re-verified during Rename operations to
- * confirm that the contact was registered on this device.
+ * This file is responsible only for serializing the HMAC message before
+ * calling the appropriate syscall.
  */
 
 #include <string.h>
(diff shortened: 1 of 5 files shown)


app_features/address_book/src/address_book_crypto.c
```diff
@@ -222,36 +144,27 @@ bool address_book_verify_group_handle(const path_bip32_t *bip32_path,
...
-
-    if (!init_hmac_ctx(bip32_path, salt, sizeof(salt) - 1, &hmac_ctx)) {
-        goto end;
-    }
-    CX_CHECK(cx_hmac_no_throw(
-        (cx_hmac_t *) &hmac_ctx, CX_LAST, gid, GID_SIZE, mac_expected, CX_SHA256_SIZE));
-    if (os_secure_memcmp(mac, mac_expected, CX_SHA256_SIZE) != 0) {
+    bool           success = false;
+    const uint8_t *gid     = group_handle;
+    const uint8_t *mac     = group_handle + GID_SIZE;
```
cedelavergne-ledger, line 241: Do we need to keep the gid and mac ocal variables here? Or does it make sense to encapsulate (and hide) the mechanism behind the syscall?

app_features/address_book/src/address_book_crypto.c
```diff
@@ -176,31 +104,25 @@ bool address_book_send_register_identity_response(const uint8_t group_handle[GRO
...
 {
-    static const uint8_t salt[]   = "AddressBook-Group";
-    cx_hmac_sha256_t     hmac_ctx = {0};
-    cx_err_t             error    = CX_INTERNAL_ERROR;
-    bool                 success  = false;
-    uint8_t             *gid      = group_handle;
-    uint8_t             *mac      = group_handle + GID_SIZE;
+    bool     success = false;
+    uint8_t *gid     = group_handle;
+    uint8_t *mac     = group_handle + GID_SIZE;
```
cedelavergne-ledger, line 211: Do we need to keep the gid and mac ocal variables here? Or does it make sense to encapsulate (and hide) the mechanism behind the syscall?
Same for the random generation: it maybe inside the syscall?

app_features/address_book/src/address_book_crypto.c
```diff
@@ -264,28 +177,31 @@ bool address_book_compute_hmac_proof(const path_bip32_t *bip32_path,
...
+    size_t  offset                                          = 0;
+    bool    success                                         = false;
+    uint8_t name_len                                        = (uint8_t) strlen(name);
 
-    if (!init_hmac_ctx(bip32_path, salt, sizeof(salt) - 1, &hmac_ctx)) {
-        goto end;
-    }
     // gid(32)
-    CX_CHECK(cx_hmac_no_throw((cx_hmac_t *) &hmac_ctx, 0, gid, GID_SIZE, NULL, 0));
+    memmove(&msg[offset], gid, GID_SIZE);
```
cedelavergne-ledger: The `msg` construction is repeated several times between the different functions below.
It would be interesting to factorize this into a static local function

cedelavergne-ledger (changes requested):

app_features/address_book/src/address_book_crypto.c
srasoamiaramanana-ledger: I'd rather keep these variables outside of the syscalls to make them generic: the purpose of the syscalls is only to derive the key and generate/verify the hmac value whatever the data provided.

app_features/address_book/src/address_book_crypto.c
srasoamiaramanana-ledger: Same comment as above.

cedelavergne-ledger (approved):

abonnaudet-ledger (approved):

PR #1650, opened 2026-07-10
Fix stack protector

mbrousset-ledger:
## Description

There were two problems in the initial stack protector implementation :
- canary was not correctly initialized because entry point was forced with `ENTRY(main);`
- during libcalls, caller's BSS was entirely wiped included canary.
The second problem could not be detected during tests as canary was not initialized.

The idea of this fix is to have the correct entry point whether building with SSP.
Also, moving canary out of bss with appropriate pic adaptation.

## Changes include

- [ ] Tests
- [ ] Documentation

## Breaking changes

## Additional comments

Remember:

### DIFF
--- a/src/os.c
+++ b/src/os.c
@@ -198,7 +198,7 @@ int bytes_to_lowercase_hex(char *out, size_t outl, const void *value, size_t len
     return 0;
 }
 
-// BSS start and end defines addresses
+// BSS wipe bounds
 extern void *_bss;
 extern void *_ebss;
 
--- a/src/stack_protector_init.S
+++ b/src/stack_protector_init.S
@@ -41,21 +41,4 @@ __wrap___stack_chk_init:
 // place the constants pool after the unconditional branch, out of the exec path
 .ltorg
 
-#else
-
-// Force the following linker error if -fstack-protector is enabled for CFLAGS
-// but not for AFLAGS:
-//
-//     ld.lld: error: duplicate symbol: __stack_chk_fail
-//
-// It ensures that __stack_chk_init is always defined (and consequently
-// __stack_chk_guard is always initialized to a random value) when SSP is
-// enabled.
-//
-// This code will never end up in an app since the link step will fail.
-.weak __stack_chk_fail
-.thumb_func
-__stack_chk_fail:
-    udf 255
-
 #endif
--- a/src/stack_protector_init.S
+++ b/src/stack_protector_init.S
@@ -7,9 +7,12 @@
 
 .section .boot.ssp_init, "ax"
 
+.extern main
 .global __wrap___stack_chk_init
 .thumb_func
 __wrap___stack_chk_init:
+    // Relies on the OS contract for arg0 (r0): 0 on a standalone launch,
+    // non-zero (the libargs_t*) on a libcall.
     // if r0 != 0, skip initialization and jump directly to main
     // (don't overwrite parent canary during a libcall)
     cmp r0, #0
@@ -31,13 +34,12 @@ __wrap___stack_chk_init:
     // restore arguments
     pop {r0-r3}
 
-    // jump to the main function
-    b 1f
-
-// skip the constants pool
-.ltorg
-
+    // canary initialized; fall through to the shared exit below
 1:
+    b main
+
+// place the constants pool after the unconditional branch, out of the exec path
+.ltorg
 
 #else
 
@@ -51,7 +53,7 @@ __wrap___stack_chk_init:
 // enabled.
 //
 // This code will never end up in an app since the link step will fail.
-.global __stack_chk_fail
+.weak __stack_chk_fail
 .thumb_func
 __stack_chk_fail:
     udf 255
--- a/target/apex_m/script.ld
+++ b/target/apex_m/script.ld
@@ -49,7 +49,7 @@ SECTIONS
     _text = .;
     _nvram_start = .;
 
-    /* ensure __stack_chk_init is always @ 0xC0D00000 */
+    /* ensure __stack_chk_init is always @ 0xC0DE0001 */
     KEEP(*(.boot.ssp_init))
     /* main; reached via b main, adjacency not required */
     *(.boot*)
@@ -133,7 +133,7 @@ SECTIONS
   } > SRAM = 0x00
 
   /* SSP: canary is written via r9 (RAM base), so guard must be first in SRAM */
-  ASSERT( __stack_chk_guard == _ram, "__stack_chk_guard must be at the RAM base (_ram)" )
+  ASSERT( __stack_chk_guard == ORIGIN(SRAM), "__stack_chk_guard must be at the RAM base" )
 
   ASSERT( (_estack - _stack) >= STACK_MIN_SIZE, "stack section too small" )
 
--- a/target/apex_m/script.ld
+++ b/target/apex_m/script.ld
@@ -33,7 +33,8 @@ PAGE_SIZE  = 512;
 STACK_MIN_SIZE = DEFINED(stack_min_size) ? stack_min_size : 1500;
 END_STACK  = ORIGIN(SRAM) + LENGTH(SRAM);
 
-ENTRY(main);
+/* __entry == __stack_chk_init if SSP is enabled, else main */
+ENTRY(__entry);
 
 SECTIONS
 {
@@ -50,7 +51,7 @@ SECTIONS
 
     /* ensure __stack_chk_init is always @ 0xC0D00000 */
     KEEP(*(.boot.ssp_init))
-    /* ensure main directly follows __stack_chk_init */
+    /* main; reached via b main, adjacency not required */
     *(.boot*)
 
     /* place the other code and rodata defined BUT nvram variables that are displaced in a r/w area */
@@ -107,10 +108,11 @@ SECTIONS
     /**
      * Place RAM uninitialized variables
      */
-    _bss = .;
+    _ram = .;
     __stack_chk_guard = .;
     PROVIDE(__stack_chk_guard = .);
     . += 4;
+    _bss = .;
     *(.bss*)
     _ebss = .;
 
@@ -129,6 +131,9 @@ SECTIONS
     PROVIDE(_estack = .);
 
   } > SRAM = 0x00
+
+  /* SSP: canary is written via r9 (RAM base), so guard must be first in SRAM */
+  ASSERT( __stack_chk_guard == _ram, "__stack_chk_guard must be at the RAM base (_ram)" )
 
   ASSERT( (_estack - _stack) >= STACK_MIN_SIZE, "stack section too small" )
 
--- a/Makefile.defines
+++ b/Makefile.defines
@@ -100,8 +100,13 @@ LDFLAGS  += -nostdlib -nodefaultlibs
 ENABLE_STACK_PROTECTOR ?= 0
 ifneq ($(ENABLE_STACK_PROTECTOR),0)
 	LDLIBS   += -Wl,--wrap=__stack_chk_fail -Wl,--wrap=__stack_chk_init
+	# Enter at __stack_chk_init (overrides ENTRY(main)) so the canary is randomized,
+	# including in sideloads; the init then reaches main via `b main`.
+	LDFLAGS  += -Wl,--defsym,__entry=__stack_chk_init
 	AFLAGS   += -fstack-protector-strong -Wno-unused-command-line-argument
 	CFLAGS   += -fstack-protector-strong
+else
+	LDFLAGS  += -Wl,--defsym,__entry=main
 endif
 
 ENABLE_LINK_TIME_OPTIMIZATION ?= 0
--- a/Makefile.rules_generic
+++ b/Makefile.rules_generic
@@ -169,15 +169,6 @@ LINK_DEPENDENCIES += $(APP_CUSTOM_LINK_DEPENDENCIES)
 $(BIN_DIR)/app.elf: $(LINK_DEPENDENCIES)
 	@echo "[LINK] $(BIN_DIR)/app.elf"
 	$(L)$(call link_cmdline,$(OBJECT_FILES) $(LDLIBS),$(BIN_DIR)/app.elf)
-ifneq ($(ENABLE_STACK_PROTECTOR),0)
-	@# Ensure __stack_chk_guard remains the very first symbol in .bss.
-	@BSS=$$(llvm-nm $(BIN_DIR)/app.elf | grep ' _bss$$' | cut -d' ' -f1); \
-	GUARD=$$(llvm-nm $(BIN_DIR)/app.elf | grep ' __stack_chk_guard$$' | cut -d' ' -f1); \
-	if [ -z "$$GUARD" ] || [ "$$GUARD" != "$$BSS" ]; then \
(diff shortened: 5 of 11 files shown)


src/os.c
```diff
@@ -198,14 +198,14 @@ int bytes_to_lowercase_hex(char *out, size_t outl, const void *value, size_t len
...
+extern void *_bss_wipe_start;
 extern void *_ebss;
 
-// This call will reset the value of the entire BSS segment
+// This call will reset the BSS segment (excluding the stack canary guard)
 void os_explicit_zero_BSS_segment(void)
 {
-    explicit_bzero(&_bss, ((uintptr_t) &_ebss) - ((uintptr_t) &_bss));
+    explicit_bzero(&_bss_wipe_start, ((uintptr_t) &_ebss) - ((uintptr_t) &_bss_wipe_start));
 }
```
fbeutin-ledger: Needs a bit more text to explain :
- _bss_wipe_start is `strictly` 1 byte after `__stack_chk_guard` which is strictly 1 byte after `_bss`
- in case of os_lib_call, we will inherit the __stack_chk_guard value of caller

target/apex_m/script.ld
```diff
@@ -33,7 +33,7 @@ PAGE_SIZE  = 512;
 STACK_MIN_SIZE = DEFINED(stack_min_size) ? stack_min_size : 1500;
 END_STACK  = ORIGIN(SRAM) + LENGTH(SRAM);
 
-ENTRY(main);
+ENTRY(_app_entry);
```
fbeutin-ledger: Maybe a small comment

fbeutin-ledger (commented):
Also needs more text in the `__stack_chk_guard` generation code to explain how it behaves with os_lib_call
Maybe also reroll if == 0 🤔

mbrousset-ledger:
`__wrap___stack_chk_init` documentation already mentions it skips canary generation during libcall

target/apex_m/script.ld
mbrousset-ledger: updated

src/os.c
mbrousset-ledger: updated

src/stack_protector_init.S
```diff
@@ -38,6 +39,7 @@ __wrap___stack_chk_init:
 .ltorg
 
 1:
+    b main
```
iartemov-ledger, line 39: Claude proposes 1/ to simplify a bit code (no 2nd jump), 2/ to update the comment about jump  and 3/ to extend the  OS contract comment at the beginning:
```
diff --git a/src/stack_protector_init.S b/src/stack_protector_init.S
index 93f649a7..c88e88be 100644
--- a/src/stack_protector_init.S
+++ b/src/stack_protector_init.S
@@ -11,6 +11,8 @@
 .global __wrap___stack_chk_init
 .thumb_func
 __wrap___stack_chk_init:
+    // Relies on the OS contract for arg0 (r0): 0 on a standalone launch,
+    // non-zero (the libargs_t*) on a libcall.
     // if r0 != 0, skip initialization and jump directly to main
     // (don't overwrite parent canary during a libcall)
     cmp r0, #0
@@ -32,15 +34,14 @@ __wrap___stack_chk_init:
     // restore arguments
     pop {r0-r3}
 
-    // jump to the main function
-    b 1f
-
-// skip the constants pool
-.ltorg
-
+    // canary initialized; fall through to the shared exit below
 1:
+    // both paths (fresh launch and libcall) continue to main
     b main
 
+// place the constants pool after the unconditional branch, out of the exec path
+.ltorg
+
 #else
 

```

iartemov-ledger (commented):
Tested with `app-bitcoin` built as:
```
make -C . -j DEBUG=1 ENABLE_STACK_PROTECTOR=1 TARGET=flex BOLOS_SDK=/tmp/my-sdk/
```
against the current SDK. 

it works with Speculos and with the real physical device:
```
echo "EF01000000" | python3 -m ledgerblue.runScript --apdu
HID => ef01000000
HID <= 4ccad0829000
<= Clear bytearray(b'L\xca\xd0\x82')
```

Need still to test it with swap.

src/stack_protector_init.S
jarevalo-ledger, line 39: IMO it's better to choose the entry-point at build-time, based on `ENABLE_STACK_PROTECTOR`  an `__app_entry` symbol can be defined to either `main` or `__wrap___stack_chk_init`, less jump.

src/os.c
```diff
@@ -198,14 +198,16 @@ int bytes_to_lowercase_hex(char *out, size_t outl, const void *value, size_t len
...
 extern void *_ebss;
 
-// This call will reset the value of the entire BSS segment
+// This call will reset the BSS segment (excluding the stack canary guard)
+// In case of an os_lib_call, callee will inherit caller's __stack_chk_guard, __stack_chk_init skips
+// canary generation during lib calls.
 void os_explicit_zero_BSS_segment(void)
 {
-    explicit_bzero(&_bss, ((uintptr_t) &_ebss) - ((uintptr_t) &_bss));
+    explicit_bzero(&_bss_wipe_start, ((uintptr_t) &_ebss) - ((uintptr_t) &_bss_wipe_start));
```
jarevalo-ledger: It let me wonder why the stack canary is in the .bss section, should be placed before.

iartemov-ledger:
Tested successfully with `app-exchange` on real Flex device.
[reproduce.txt](https://github.com/user-attachments/files/30081481/reproduce.txt)

src/stack_protector_init.S
iartemov-ledger: > at build-time, based on ENABLE_STACK_PROTECTOR

This is what is already done in the PR - if `ENABLE_STACK_PROTECTOR` the entry point is overridden to `__wrap___stack_chk_init`. 
And it's needed for Speculos as it starts apps at `ENTRY(main) -> e_entry`
Do I understand correctly the the OS starts a side-loaded apps also at `ENTRY(main) -> e_entry`, but trusted apps at the start of the app's code region?

Also to do the explicit jump seems to be less fragile than to rely on  fall-through mechanism assuming that `main()` goes always after `__stack_chk_init()`.

And one jump costs nothing :)

mbrousset-ledger: Indeed, during tests of the target branch i observed the canary was never initialized when starting a sideloaded app (canary == `0x00000000`), denoting entry point to be main (while building with `ENABLE_STACK_PROTECTOR=1`), while this was not case in trusted apps where the canary was correctly initialized.

src/os.c
mbrousset-ledger: correct, will fix

src/stack_protector_init.S
jarevalo-ledger: After check and as discussed, applications, side-loaded or not,  will always start at the address of the entry-point set by the linker-script. As mentioned, Speculos mimics this.

src/os.c
mbrousset-ledger: updated

src/stack_protector_init.S
mbrousset-ledger: fixed

mbrousset-ledger: updated

mbrousset-ledger:
Currently, CI is failing because of rust apps as this PR adds a `_ram` symbol for pic relocation. the rust sdk needs it as well, this is done [here](https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/416) and requires rust sdk release cc @yogh333.

src/stack_protector_init.S
```diff
@@ -51,7 +54,7 @@ __wrap___stack_chk_init:
 // enabled.
 //
 // This code will never end up in an app since the link step will fail.
-.global __stack_chk_fail
+.weak __stack_chk_fail
```
iartemov-ledger: Why was it changed to `weak`? 
I did an artificial experiment - added to `Makefile.defines`:
```
diff --git a/Makefile.defines b/Makefile.defines
index 842c86f6..22388b86 100644
--- a/Makefile.defines
+++ b/Makefile.defines
@@ -107,6 +107,7 @@ ifneq ($(ENABLE_STACK_PROTECTOR),0)
        CFLAGS   += -fstack-protector-strong
 else
        LDFLAGS  += -Wl,--defsym,__entry=main
+       CFLAGS   += -fstack-protector-strong
 endif
```
(only for `CFLAGS`)

and I got:
```
[LINK] build/flex/bin/app.elf
ld.lld: error: undefined symbol: _exit
>>> referenced by libc_a-stack_protector.o:(__stack_chk_fail) in archive /tmp/my-sdk//arch/st33k1/lib/libc.a

ld.lld: error: undefined symbol: _write
>>> referenced by libc_a-writer.o:(_write_r) in archive /tmp/my-sdk//arch/st33k1/lib/libc.a
>>> did you mean: write
>>> defined in: /tmp/my-sdk//arch/st33k1/lib/libc.a(libc_a-syswrite.o)

ld.lld: error: undefined symbol: _kill
>>> referenced by libc_a-signalr.o:(_kill_r) in archive /tmp/my-sdk//arch/st33k1/lib/libc.a

ld.lld: error: undefined symbol: _getpid
>>> referenced by libc_a-signalr.o:(.text._getpid_r+0x0) in archive /tmp/my-sdk//arch/st33k1/lib/libc.a
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [/tmp/my-sdk//Makefile.rules_generic:171: build/flex/bin/app.elf] Error 1
make: Leaving directory '/app'
```
Not `a " duplicate symbol: __stack_chk_fail`.

If I revert to `.global` I do get:
```
[LINK] build/flex/bin/app.elf
ld.lld: error: duplicate symbol: __stack_chk_fail
>>> defined at build/flex/obj/sdk/src/stack_protector_init.o:(.text+0x1)
>>> defined at libc_a-stack_protector.o:(.text.__stack_chk_fail+0x1) in archive /tmp/my-sdk//arch/st33k1/lib/libc.a
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [/tmp/my-sdk//Makefile.rules_generic:171: build/flex/bin/app.elf] Error 1
make: Leaving directory '/app
```

So if we want to keep the check and the match to the comment then we need to restore the `.global`.

target/apex_m/script.ld
```diff
@@ -33,7 +33,8 @@ PAGE_SIZE  = 512;
 STACK_MIN_SIZE = DEFINED(stack_min_size) ? stack_min_size : 1500;
 END_STACK  = ORIGIN(SRAM) + LENGTH(SRAM);
 
-ENTRY(main);
+/* __entry == __stack_chk_init if SSP is enabled, else main */
+ENTRY(__entry);
 
 SECTIONS
```
iartemov-ledger, line 39: We might want to update the `/* ensure main directly follows __stack_chk_init */` comment below to something like `/* main; reached via `b main`, adjacency not required */` ?

target/apex_m/script.ld
```diff
@@ -107,10 +108,11 @@ SECTIONS
     /**
      * Place RAM uninitialized variables
      */
-    _bss = .;
+    _ram = .;
```
iartemov-ledger, line 111: Instead of removed `__stack_chk_guard` checking in `Makefile.rules_generic` we can add an assert in  the linker script:
```
  /* SSP: canary is written via r9 (RAM base), so guard must be first in SRAM */
  ASSERT( __stack_chk_guard == _ram, "__stack_chk_guard must be at the RAM base (_ram)" )
```

src/stack_protector_init.S
```diff
@@ -31,14 +34,14 @@ __wrap___stack_chk_init:
     // restore arguments
     pop {r0-r3}
 
-    // jump to the main function
-    b 1f
+    // canary initialized; fall through to the shared exit below
+1:
+    b main
 
+// place the constants pool after the unconditional branch, out of the exec path
```
iartemov-ledger, line 41: Bit redundant comments?

iartemov-ledger (commented):
Few comments. 
Also I re-an my tests (see `reproduce.txt`) and it works (app-bitcoin tested with Speculos and on real device, exchange test on real device) - the canary is updated on each run. .

src/stack_protector_init.S
mbrousset-ledger: updated

target/apex_m/script.ld
mbrousset-ledger: good point, updated

target/apex_m/script.ld
mbrousset-ledger: good point, updated

src/stack_protector_init.S
mbrousset-ledger: For your first batch of errors : if  you define -fstack-protector-strong you also need to redirect it to wrapped symbol, libc stack protector implementation uses undefined libc functions hence the errors you saw.

weak is needed because so it works also with LTO.

iartemov-ledger: My point was that initially the `.global`  served to ensure the consistency check.

bigspider: I unresolved the comment for visibility.

This was pointed out by local AI review, and I agree that it at the very least contradicts the comment above. So if it cannot be kept strong, then the comment should be updated accordingly.

target/apex_m/script.ld
```diff
@@ -130,6 +132,9 @@ SECTIONS
 
   } > SRAM = 0x00
 
+  /* SSP: canary is written via r9 (RAM base), so guard must be first in SRAM */
+  ASSERT( __stack_chk_guard == _ram, "__stack_chk_guard must be at the RAM base (_ram)" )
```
bigspider: The assertion only protects from adding variables between `_ram` and `__stack_chk_guard = .;`, but nothing prevents from adding a new section in RAM before `.bss`, or other variables before `_ram`.

A stronger assertion might use `ORIGIN(SRAM)`:

```
ASSERT( __stack_chk_guard == ORIGIN(SRAM), "__stack_chk_guard must be at the RAM base" )
```

(same of course for all the other copies)

target/apex_m/script.ld
```diff
@@ -50,7 +51,7 @@ SECTIONS
 
     /* ensure __stack_chk_init is always @ 0xC0D00000 */
```
bigspider: nit (preexisting): `0xc0de0000` seems to be the right number (address of `FLASH`)

src/stack_protector_init.S
mbrousset-ledger: Good point, the initial purpose of this trap was to generate a linker error if `AFLAGS` and `CFLAGS` were misaligned. But they are controlled together through `ENABLE_STACK_PROTECTOR` now, so i think it is safe to remove this trap.

target/apex_m/script.ld
mbrousset-ledger: Updated

target/apex_m/script.ld
mbrousset-ledger: good remark, thanks  for seeing this. Updated.

src/stack_protector_init.S
mbrousset-ledger: removed

bboilot-ledger (commented):
LGTM!

bboilot-ledger (approved):

iartemov-ledger (approved):
Tested it with `app-bitcoin` changing the canary on each new run on Flex, and with  `app-exchange` test showing the `app-bitcoin` canary change.

PR #1661, opened 2026-07-29
script.ld and plugin_script.ld refactoring

iartemov-ledger:
#### Per-target differences resolved by this PR

| Target | SRAM | PAGE\_SIZE | FLASH/SRAM fill `= 0x00` | `.storage_section` | Debug section |
|--------|------|-----------|--------------------------|---------------------|---------------|
| apex\_m | 40K | 512 | ✅ | ~~missing~~ → ✅ fixed | `DEBUG (NOLOAD)` |
| apex\_p | 40K | 512 | ✅ | ~~missing~~ → ✅ fixed | `DEBUG (NOLOAD)` |
| flex   | 36K | 512 | ✅ | ✅ | `DEBUG (NOLOAD)` |
| nanos2 | 40K | 512 | ✅ | ✅ | `DEBUG (NOLOAD)` |
| nanox  | 28K | 256 | ~~missing~~ → ✅ fixed | ✅ | ~~`/DISCARD/`~~ → `DEBUG (NOLOAD)` fixed |
| stax   | 36K | 512 | ✅ | ✅ | `DEBUG (NOLOAD)` |

The MEMORY block and `PAGE_SIZE` remain per-target in each `script.ld`.
Everything else moves to `target/common.ld` and is shared via `INCLUDE ../common.ld`.

##### Behaviour changes vs. the old scripts

- **apex\_m, apex\_p** — gained `*(.storage_section)` (was missing by oversight).
- **nanox** — debug section changed from `/DISCARD/` (truly removes sections from the ELF)
  to `DEBUG (NOLOAD) : { … } > DISCARD` (keeps sections in ELF at a dummy VMA, not loaded
  at runtime). No runtime difference on hardware; aligns with the other five targets.
- **nanox** — `} > FLASH` and `} > SRAM` now carry the `= 0x00` fill byte, matching all
  other targets. Uninitialised flash/SRAM bytes in the output image are now explicitly
  zeroed rather than left at the toolchain default.

**The INCLUDE approach adopted in this PR seems to me better than the one used in https://github.com/LedgerHQ/ledger-secure-sdk/pull/398. So this PR is effectively overriding it and will close it.** 

### plugin_script.ld

- [x] only one `target/plugin_script.ld` remains for all targets
- [x]  `ld.lld: error: .bss section must be empty for plugin` fixed

### Result binary tests with different apps

| App | Target | Script used | Result |
|-----|--------|-------------|--------|
| app-bitcoin-new    | apex_p | `script.ld`        | ✅ identical |
| app-bitcoin-new    | nanox  | `script.ld`        | ⚠️ fill change — see below |
| app-bitcoin-new    | nanos2 | `script.ld`        | ✅ identical |
| app-bitcoin-new    | flex   | `script.ld`        | ✅ identical |
| app-bitcoin-new    | stax   | `script.ld`        | ✅ identical |
| app-plugin-uniswap | apex_p | `plugin_script.ld` | ✅ identical |
| app-plugin-uniswap | nanox  | `plugin_script.ld` | ⚠️ fill change — see below |
| app-plugin-uniswap | nanos2 | `plugin_script.ld` | ✅ identical |
| app-plugin-uniswap | flex   | `plugin_script.ld` | ✅ identical |
| app-plugin-uniswap | stax   | `plugin_script.ld` | ✅ identical |
| app-boilerplate    | apex_p | `script.ld`        | ✅ identical |
| app-boilerplate    | nanox  | `script.ld`        | ⚠️ fill change — see below |
| app-boilerplate    | nanos2 | `script.ld`        | ✅ identical |
| app-boilerplate    | flex   | `script.ld`        | ✅ identical |
| app-boilerplate    | stax   | `script.ld`        | ✅ identical |

**15 target/app combinations tested: 12 identical, 3 differ (all on nanox, same root cause).**

#### nanox difference explained

**Root cause:** `target/nanox/script.ld` in `master` used `} > FLASH` (no fill specifier).
`common.ld` in `merge_script_ld` uses `} > FLASH = 0x00`.

When no section fill is specified, LLVM/Clang's compiler-generated alignment padding bytes (0xD4D4 = Thumb-16 BMI #-84, a real conditional branch) are preserved in the output. With = 0x00, the linker overrides those gaps with 0x00 0x00 (MOVS r0, r0, a no-op), which is safer since a stray branch is worse than a stray no-op if execution ever reaches padding.

The gaps are never executed (they are alignment padding between functions/sections at
4-byte or `PAGE_SIZE` boundaries). The change is **functionally inert at runtime**.

| App | Differing hex records | Scope |
|-----|----------------------:|-------|
| app-bitcoin-new    | 184 records (×16 B ≈ 2.9 kB) | alignment gaps in `.text` |
| app-plugin-uniswap |  28 records (×16 B = 448 B)  | alignment gaps in `.text` |
| app-boilerplate    | 116 records (×16 B ≈ 1.9 kB) | alignment gaps in `.text` |

**This is an intentional standardisation**: all other targets (apex_p, nanos2, flex, stax)
already had `= 0x00` fill in master or had lld produce zero fill by default; nanox was
the odd one out. The `merge_script_ld` branch makes the fill uniform across all targets

## Changes include

- [ ] Tests
- [ ] Documentation

Remember:


iartemov-ledger:
@jarevalo-ledger, it would be good to check that the change does not break the OS build ;)

cedelavergne-ledger (approved):

cedelavergne-ledger (approved):

jarevalo-ledger:
> @jarevalo-ledger, it would be good to check that the change does not break the OS build ;)

It's ok :+1:

jarevalo-ledger (approved):

iartemov-ledger:
Just few notes following offline disucssions with @jarevalo-ledger 

> used linker

[Here](https://github.com/LedgerHQ/ledger-secure-sdk/blob/merge_script_ld/Makefile.defines#L95) with LDFLAGS += -fuse-ld=lld we override the linker by default by `lld`.

> The fill in value

Initial comment about 0xD4 fill was incorrect - see the corrected one in the description

> What you'd actually want for defense-in-depth:
> 0xDE00 = BKPT #0 (breakpoint → debug/hard fault) or 0xDEFF — both cause an immediate fault if accidentally executed
> . That would be the principled choice. But switching to = 0xDEFF or similar:
> Would require validating the new fill is a valid undefined/trap encoding for all Thumb-2 variants used across the 6 targets
> Would produce another binary diff vs. master across all targets (not just nanox)
> Goes beyond what this PR sets out to do
> Conclusion: Keep = 0x00. The five other targets already had it, the binary tests validated it, and 0x00 (NOP) is safer than 0xD4 (real conditional branch). If the team later wants defensive padding uniformly, that's a separate PR targeting all fill expressions with an explicit trap encoding.

mbrousset-ledger:
@iartemov-ledger I think it would have been good to `git mv` an older linker script to `common.ld` than removing all older linker scripts and creating a new file, because this loses all linker script history.

Using `git mv` preserves history

PR #1666, opened 2026-08-03
Add address book tests

cedelavergne-ledger:
## Description

Add tests for Address Book feature:
- Fuzzing
- Unit-Tests (+ refactor the README files)

Bonus:
- Fix & optimise PIC definition to avoid compilation warnings in UT
- Fix compilation warnings in `lib_cxng/src/cx_mldsa_XXX`

## Changes include

- [X] Tests
- [ ] Documentation


mbrousset-ledger:
could be nice to have unit tests using latest CMake / Unity / CMock framework

lib_cxng/src/cx_mldsa_util.h
```diff
@@ -33,24 +33,24 @@ void MLDSA_UTIL_shake128(uint8_t *out, size_t outlen, const uint8_t *in, size_t
...
-                             const uint8_t *in1,
-                             size_t         in1len,
-                             const uint8_t *in2,
-                             size_t         in2len);
+cx_err_t MLDSA_UTIL_shake256_two(uint8_t       *out,
+                                 size_t         outlen,
+                                 const uint8_t *in1,
+                                 size_t         in1len,
+                                 const uint8_t *in2,
+                                 size_t         in2len);
```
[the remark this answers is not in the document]
cedelavergne-ledger: Fixed

fuzzing/sdk-fuzz/harness/fuzz_address_book.c
```diff
@@ -0,0 +1,257 @@
...
+    {0x02, 1,  1 }, /* TAG_STRUCTURE_VERSION */
+    {0xf0, 1,  32}, /* TAG_CONTACT_NAME      */
+    {0xf1, 0,  32}, /* TAG_SCOPE (optional)  */
+    {0xf2, 1,  80}, /* TAG_ACCOUNT_IDENTIFIER */
+    {0x69, 1,  41}, /* TAG_DERIVATION_PATH   */
+    {0x23, 1,  8 }, /* TAG_CHAIN_ID          */
+    {0x51, 1,  1 }, /* TAG_BLOCKCHAIN_FAMILY */
+    {0xf6, 64, 64}, /* TAG_GROUP_HANDLE (optional) */
+    {0x29, 32, 32}, /* TAG_HMAC_PROOF (optional)   */
+};
```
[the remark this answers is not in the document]
cedelavergne-ledger: Fixed

fuzzing/include/fuzz_mutator.h
```diff
@@ -123,10 +123,12 @@ static size_t fuzz_mutate_post_prefix(uint8_t *data, size_t size, size_t max_siz
     return LLVMFuzzerMutate(data, size, max_size);
 }
 
+#ifdef FUZZ_INJECT_TOKEN
```
iartemov-ledger, line 126: What is this define, who defines it?

unit-tests/address_book/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+    NAME         test_address_book_common
+    SOURCES      ${AB_SOURCES_COMMON}
+    INCLUDE_DIRS ${AB_INCLUDE_DIRS}
+    COMPILE_DEFS ${AB_COMPILE_DEFS}
+)
+
+ledger_unit_tests_add_test(
+    NAME         test_address_book_provide_contact
+    SOURCES      ${AB_SOURCES_COMMON}
+                 ${SDK_DIR}/app_features/address_book/src/address_book_crypto.c
```
iartemov-ledger, line 46: It seems that at least `address_book_crypto.c` coverage can be increased compared to others (currently is at `47.7%`)?

fuzzing/sdk-fuzz/harness/fuzz_address_book.c
```diff
@@ -0,0 +1,257 @@
+/* Absolution dispatcher for addr_book_handle_apdu (all sub-commands). */
```
iartemov-ledger, line 1: The fuzzing coverage seems to be relatively low:

**23.8 % lines (420/1764) · 42.1 % functions (56/133) · 20.3 % branches (83/408) · 0 crashes**

| File | Lines | Functions | Branches |
|---|---|---|---|
| `address_book.c` | **75.6 %** | 100 % | 76.2 % |
| `address_book_common.c` | 65.2 % | 57.1 % | 87.5 % |
| `ledger_account_provide_contact.c` | 39.7 % | 80.0 % | 19.2 % |
| `identity_register.c` | 31.5 % | 68.8 % | 20.4 % |
| `identity_provide_contact.c` | 25.0 % | 50.0 % | 11.1 % |
| `identity_edit_identifier.c` | 19.9 % | 38.9 % | 10.9 % |
| `ledger_account_register.c` | 18.4 % | 36.4 % | 7.1 % |
| `ledger_account_edit.c` | 18.2 % | 38.5 % | 8.8 % |
| `identity_edit_scope.c` | 15.1 % | 27.8 % | 8.3 % |
| `identity_edit_contact_name.c` | 15.1 % | 23.1 % | 10.7 % |
| `address_book_crypto.c` | **0.00 %** | 0.00 % | 0.00 % |

unit-tests/address_book/io.h
```diff
@@ -0,0 +1,29 @@
+/* io.h — minimal unit-test stub.
```
mbrousset-ledger: UT should use original `io.h` if possible, and use mock instead of actual implementation

unit-tests/address_book/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+ledger_unit_tests_add_test(
+    NAME         test_address_book_common
+    SOURCES      ${AB_SOURCES_COMMON}
+    INCLUDE_DIRS ${AB_INCLUDE_DIRS}
+    COMPILE_DEFS ${AB_COMPILE_DEFS}
+)
+
+ledger_unit_tests_add_test(
+    NAME         test_address_book_provide_contact
+    SOURCES      ${AB_SOURCES_COMMON}
```
mbrousset-ledger, line 44:
```suggestion
    MOCK_HEADERS ${SDK_DIR}/io/io.h
    SOURCES      ${AB_SOURCES_COMMON}
```

instead of having custom io

unit-tests/address_book/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+    ${SDK_DIR}/lib_standard_app/buffer.c
+    ${SDK_DIR}/lib_standard_app/read.c
+    ${SDK_DIR}/lib_standard_app/write.c
+    ${SDK_DIR}/lib_standard_app/varint.c
+    ${SDK_DIR}/app_features/address_book/src/address_book_common.c
+)
+
+set(AB_INCLUDE_DIRS
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    ${SDK_DIR}/app_features/address_book/include
```
mbrousset-ledger, line 21:
```suggestion
    ${SDK_DIR}/io/include
    ${SDK_DIR}/app_features/address_book/include
```

unit-tests/address_book/stubs.c
```diff
@@ -0,0 +1,140 @@
+/*****************************************************************************
```
mbrousset-ledger: mock should be build artifacts and should not be versioned

include/os_pic.h
```diff
@@ -1,19 +1,15 @@
...
+// Position-independent code reference
+// Function that align the dereferenced value in a rom struct to use it depending on the execution
+// address. Can be used even if code is executing at the same place where it had been linked.
+#ifdef OS_UNIT_TEST
+#define PIC(x) (x)
+#elif defined(HAVE_BOLOS) && !defined(BOLOS_OS_UPGRADER_APP)
+#define PIC(x) pic_shared((const void *) x)
+void *pic_shared(const void *linked_address);
+#else
+#define PIC(x) pic((void *) x)
```
iartemov-ledger, line 13: `#ifndef PIC` dropped seems to give a warning for `bagl.c` compilation as it is the consumer:
```
lib_bagl/src/bagl.c:19-22, before including bagl.h → os.h → os_pic.h
```

unit-tests/address_book/test_address_book_provide_contact.c
```diff
@@ -0,0 +1,732 @@
...
+    (void) account;
+    return g_mock_provide_la_result;
+}
+
+/* ── TLV builder ─────────────────────────────────────────────────────────── */
+/* Tags < 0x80:  DER short form, 1 byte.                                      */
+/* Tags 0x80..0xFF: DER long form, 2 bytes [0x81, tag].                       */
+/* All payload field sizes are < 0x80, so lengths are always 1 byte.          */
+
+static void tlv_append(uint8_t *buf, size_t *off, uint32_t tag, const uint8_t *val, uint8_t vlen)
```
iartemov-ledger: `tlv_append` / `tlv_u8` and the `BIP32_ETH_PATH` / `DUMMY_IDENTIFIER` / `ZERO_32` /
`ZERO_64` / `ETH_CHAIN_ID_1` constants are copy-pasted between
`test_address_book_provide_contact.c:84-121` and `test_address_book_ui_flows.c:143-174`.
Can they be factorized in a common file?

fuzzing/sdk-fuzz/harness/fuzz_address_book.c
```diff
@@ -0,0 +1,257 @@
...
+     * Clamp to the address book's maximum chunk size so the VLA stays bounded. */
+#define AB_MAX_CHUNK (OS_IO_SEPH_BUFFER_SIZE - 3 - 5)
+    size_t  payload_len = fuzz_tail_len > AB_MAX_CHUNK ? AB_MAX_CHUNK : fuzz_tail_len;
+    size_t  buf_size    = 2 + payload_len;
+    uint8_t buf[2 + AB_MAX_CHUNK];
+    buf[0] = (uint8_t) (payload_len >> 8);
+    buf[1] = (uint8_t) (payload_len & 0xFF);
+    memcpy(buf + 2, fuzz_tail_ptr, payload_len);
+
+    addr_book_handle_apdu(buf, buf_size, c->p1, 0x00);
```
iartemov-ledger: P2 is hard-codded as 0x00. Consequences:

- `.p2_max = 0x80` (line 218) is dead
- The comment on line 216 ("P2 toggles between 0x00 (first chunk) and 0x80
  (continuation)") describes something the code does not do

**Fix:** pass `c->p2` through to `addr_book_handle_apdu`, and when it selects
continuation, split `fuzz_tail` into a first chunk with a *deliberately inconsistent*
header plus one or more continuation chunks.

unit-tests/address_book/stubs.c
```diff
@@ -0,0 +1,140 @@
...
+{
+    (void) rdatalist;
+    (void) count;
+    (void) sw;
+    return 0;
+}
+
+/* ── NBGL stubs ──────────────────────────────────────────────────────────── */
+
+void nbgl_useCaseReviewLight(nbgl_operationType_t              operationType,
```
iartemov-ledger: Because `nbgl_useCaseReviewLight` drops its callback, `review_choice` and
`build_and_send_response` are **never called in any of the six flows**, and neither is
anything downstream:

- `address_book_crypto.c`: `address_book_compute_hmac_proof`,
  `address_book_compute_hmac_proof_ledger_account`, `address_book_compute_hmac_rest`,
  `address_book_generate_group_handle`, `address_book_send_hmac_proof`,
  `address_book_send_register_identity_response` — **all 0 calls**
- `address_book_common.c`: `address_book_finalize_review`,
  `address_book_handle_review_rejected` — **0 calls**

That is why `address_book_crypto.c` sits at 47.7 % despite being compiled into two suites.

**Fix (cheap):** exactly what the fuzz harness already does for
`display_register_ledger_account_review` — have the stub invoke `callback(true)` /
`callback(false)` under a `g_mock_review_choice` global. That single change adds the
reject path *and* most of the crypto layer.

unit-tests/address_book/test_address_book_provide_contact.c
```diff
@@ -0,0 +1,732 @@
...
+                                       false); /* no hmac_rest */
+    TEST_ASSERT_EQUAL_INT(SWO_INCORRECT_DATA, provide_contact(buf, len));
+}
+
+static void test_pc_hmac_verify_fails(void)
+{
+    uint8_t buf[512];
+    size_t  len = build_valid_provide_contact_btc(buf, sizeof(buf));
+
+    g_mock_hmac_verify_result = false;
```
iartemov-ledger: **Mock state is restored at the wrong place; a single failure cascades**

Every test does `g_mock_* = false; TEST_ASSERT_…; g_mock_* = true;`. Unity's `RUN_TEST`
longjmps out of the test body on a failed assert, so the restore line is skipped: if
`test_pc_hmac_verify_fails` ever fails, `g_mock_hmac_verify_result` stays `false` and
every subsequent test in the file fails too, hiding the real culprit.

**Fix:** reset all `g_mock_*` in `setUp()` — both files, ~10 sites.

unit-tests/README.md
```diff
@@ -0,0 +1,54 @@
+# Unit Tests
+
+## Prerequisites
+
+- CMake >= 3.10
+- CMocka >= 1.1.5
+- lcov >= 1.14 (for code coverage)
```
iartemov-ledger, line 9: **The prerequisites are incomplete for the new framework**

`unit-tests/README.md` lists only "CMocka >= 1.1.5", but `address_book/` uses
**Unity + CMock** via `cmake/LedgerUT.cmake`, which needs **Ruby** - precisely why this
PR adds `apt-get install -y ruby` to the workflow. It should say: Ruby required;
Unity/CMock fetched from GitHub at configure time (so the build needs network); CMocka
still required for the older suites (`lib_alloc`, `lib_standard_app`, …).

unit-tests/address_book/io.h
cedelavergne-ledger: Fixed

unit-tests/address_book/CMakeLists.txt
cedelavergne-ledger: Fixed

unit-tests/address_book/CMakeLists.txt
cedelavergne-ledger: Not needed, because the lib_standard_app/io.h is already included in `UT_INCLUDE_DIRS_SDK`

unit-tests/address_book/stubs.c
cedelavergne-ledger: Done, renamed

unit-tests/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Code Coverage
+
+From any suite directory:
+
+```bash
+../gen_coverage.sh
+```
+
+Outputs `coverage.total` and `coverage/index.html`.
```
iartemov-ledger: Does not seem to be completely correct (to be checked).
 `gen_coverage.sh` produces
`coverage.info` (plus `coverage.base` / `coverage.capture`). No `coverage.total` anywhere.

lib_cxng/src/cx_mldsa_util.c
```diff
@@ -34,44 +34,64 @@ void MLDSA_UTIL_shake128(uint8_t *out, size_t outlen, const uint8_t *in, size_t
...
-    uint8_t buf[256] = {0};
-    size_t  total    = in1len + in2len;
+    uint8_t  buf[256] = {0};
+    size_t   total    = in1len + in2len;
+    cx_err_t error;
 
     if (total <= sizeof(buf)) {
         (void) memcpy(buf, in1, in1len);
         (void) memcpy(buf + in1len, in2, in2len);
         cx_shake256_hash(buf, total, out, outlen);
```
iartemov-ledger: Should be `error = cx_shake256_hash(...)` with the `error` propagated.

lib_cxng/src/cx_mldsa_util.c
```diff
@@ -34,44 +34,64 @@ void MLDSA_UTIL_shake128(uint8_t *out, size_t outlen, const uint8_t *in, size_t
...
+        return error;
+    }
+    error = cx_hash_no_throw((cx_hash_t *) &ctx, 0, in2, in2len, NULL, 0);
+    if (error != CX_OK) {
+        return error;
+    }
+    return cx_hash_no_throw((cx_hash_t *) &ctx, CX_LAST, in3, in3len, out, outlen);
 }
 
 void MLDSA_UTIL_shake128_seed_nonce(uint8_t       *out,
```
iartemov-ledger, line 96: `MLDSA_UTIL_shake128` and `MLDSA_UTIL_shake256` do not propagate `cx_err_t`-returning calls.

iartemov-ledger (commented):
Several comments. 
generic remark: in a next evolution of this we should find a way to mutualise the UT and the fuzzing code.

fuzzing/include/fuzz_mutator.h
cedelavergne-ledger: I don't know. Here, I just fixed the compilation warning, because `fuzz_should_inject_token` was "defined but not used". The flag is the same already used in this file

include/os_pic.h
cedelavergne-ledger: Fixed

unit-tests/address_book/test_address_book_provide_contact.c
cedelavergne-ledger: Added a common helper

unit-tests/address_book/stubs.c
cedelavergne-ledger: Removed, and replaces by generated mocks from cmake

unit-tests/address_book/test_address_book_provide_contact.c
cedelavergne-ledger: Done

unit-tests/README.md
cedelavergne-ledger: Updated.
Will change also in the next PR to migtate all UT to the new framework

unit-tests/README.md
cedelavergne-ledger: Updated.

fuzzing/sdk-fuzz/harness/fuzz_address_book.c
cedelavergne-ledger: Improved and fixed

lib_cxng/src/cx_mldsa_util.c
cedelavergne-ledger: Should be addressed in a separate PR

lib_cxng/src/cx_mldsa_util.c
cedelavergne-ledger: Fixed

unit-tests/address_book/CMakeLists.txt
```diff
@@ -0,0 +1,94 @@
+cmake_minimum_required(VERSION 3.10)
+project(unit_tests_address_book LANGUAGES C)
+
+set(SDK_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../..)
```
mbrousset-ledger, line 4: note : `SDK_DIR` is  temporary as the main build system is not cmake yet, this should be replaced with `CMAKE_SOURCE_DIR` eventually

mbrousset-ledger (approved):
Approving but leaving a general comment : to comply with pure unit-tests approach, there should be one test binary per compilation unit. Dependencies are then mocked for each binary, ensuring only the logic of the tested compilation unit is tested.

unit-tests/address_book/CMakeLists.txt
iartemov-ledger: Unit tests coverage for `app_features/address_book/src/`

| | Round 1 | Round 2 |
|---|---|---|
| **Lines** | 70.3 % (682/970) | **87.7 % (851/970)** |
| **Functions** | 83.2 % (124/149) | **96.6 % (144/149)** |
| Branches | no data | no data (see §5) |

| File | R1 lines | R2 lines | R2 functions |
|---|---|---|---|
| `address_book_crypto.c` | 47.7 % | **87.9 %** | **100 %** |
| `address_book_common.c` | 74.4 % | 94.9 % | 100 % |
| `ledger_account_provide_contact.c` | 96.4 % | 94.5 % | 100 % |
| `ledger_account_register.c` | 71.0 % | 91.9 % | 92.3 % |
| `identity_edit_contact_name.c` | 75.0 % | 91.7 % | 100 % |
| `identity_provide_contact.c` | 91.5 % | 91.5 % | 100 % |
| `ledger_account_edit.c` | 72.7 % | 89.6 % | 93.3 % |
| `identity_edit_scope.c` | 72.4 % | 87.0 % | 95.0 % |
| `identity_edit_identifier.c` | 70.4 % | 86.1 % | 95.0 % |
| `identity_register.c` | 62.8 % | 84.3 % | 94.4 % |
| `address_book.c` | 71.4 % | **71.4 %** (unchanged) | 100 % |

All 4 suites build warning-free (serial build) and pass.

**Only one function in the whole module is still never called:** `handle_chain_id`, in all
five register/edit flows.

fuzzing/sdk-fuzz/generate-seeds-address-book.py
```diff
@@ -0,0 +1,366 @@
...
+
+def tlv_raw(tag: int, data: bytes) -> bytes:
+    return tlv(tag, data)
+
+
+# ── Seed-file writer ─────────────────────────────────────────────────────────
+
+def write_seed(output_dir: str, name: str, p1: int, payload: bytes) -> None:
+    """Write a seed file.  payload = raw TLV bytes (no length header needed)."""
+    header = bytes([0x00, 0x00, p1, 0x00])  # ctrl=0, cmd=0, P1, P2=single-chunk
```
iartemov-ledger, line 146: #### [High] The custom seed generator emits seeds that are all discarded

[`fuzzing/sdk-fuzz/generate-seeds-address-book.py`](https://github.com/LedgerHQ/ledger-secure-sdk/blob/cev/address_book_tests/fuzzing/sdk-fuzz/generate-seeds-address-book.py)
writes a 4-byte header plus the TLV payload and **never prepends the Absolution globals
prefix**:

```python
def write_seed(output_dir: str, name: str, p1: int, payload: bytes) -> None:
    """Write a seed file.  payload = raw TLV bytes (no length header needed)."""
    header = bytes([0x00, 0x00, p1, 0x00])  # ctrl=0, cmd=0, P1, P2=single-chunk
    data   = header + payload
```

The campaign then filters every one of them — `app-campaign.sh:355` sets
`min_input_size=$((prefix_size + 4))`:

```
--- Custom seeds ---
  wrote 'register_identity_btc_new_group'      ( 74 bytes, P1=0x01)
  wrote 'register_identity_eth_existing_group' (181 bytes, P1=0x01)
  wrote 'edit_identifier_eth'                  (237 bytes, P1=0x03)
  wrote 'register_ledger_account_btc'          ( 46 bytes, P1=0x11)
  ...
  Filtered 15 corpus files smaller than 1187 bytes
```

All 15 hand-built valid-TLV seeds — the entire point of the new generator, and the fix for
round-1 finding #3 — are dropped before the fuzzer starts. Only the 32 generic seeds
(correctly sized at 1187/1439 bytes) survive.

This is deterministic arithmetic on file sizes, visible in the campaign log, and independent
of my toolchain: it will behave identically in CI and ClusterFuzzLite.

**Fix:** mirror what `generate-seed-corpus-generic.py` already does —

```python
from fuzz_seed_utils import (resolve_prefix_size, resolve_seed_prefix,
                             make_prefix_with_ctrl, parse_layout_header)

prefix_size = resolve_prefix_size(fuzzer_name="fuzz_address_book")
base_prefix = resolve_seed_prefix(prefix_size)
prefix      = make_prefix_with_ctrl(base_prefix, CTRL_OFF, ctrl)
data

File: unit-tests/address_book/CMakeLists.txt [hunk 21]
```diff
@@ -0,0 +1,94 @@
...
+
+set(AB_COMPILE_DEFS
+    ${UT_COMPILE_DEFS_SDK}
+    HAVE_ADDRESS_BOOK
+    HAVE_ADDRESS_BOOK_LEDGER_ACCOUNT
+    APPNAME="UnitTests"
+    WITH_STDIO
+)
+
+ledger_unit_tests_add_test(
```
iartemov-ledger, line 33: ####  [Medium] `make -j` races on the CMock-generated sources

`ledger_unit_tests_add_test()` writes every generated mock to one directory shared by the
whole suite (`${CMAKE_CURRENT_BINARY_DIR}/mocks/Mock<header>.c`) and adds it to each
consuming executable's source list. When several targets mock the same header, the Makefile
generator emits an identical generation rule into every consuming target's `build.make`, and
`make -j` runs those rules concurrently on the same output file.

All seven mocks in the address-book suite are shared, so there are seven race windows:

| Generated mock | Targets emitting a rule for it |
|---|---|
| `Mockio.c` | 4 |
| `Mocknbgl_use_case.c` | 3 |
| `Mockos_utils.c` | 3 |
| `Mocklcx_rng.c` | 2 |
| `Mockledger_assert_internals.c` | 2 |
| `Mockos_address_book.c` | 2 |
| `Mockos_pic.c` | 2 |

Two of five clean `-j8` builds fail:

```
/usr/lib/ruby/3.0.0/fileutils.rb:551:in `rename': No such file or directory
  @ rb_file_s_rename - (build/mocks/Mockio.h.new, build/mocks/Mockio.h) (Errno::ENOENT)
make[2]: *** [CMakeFiles/test_address_book_ui_flows.dir/build.make:75: mocks/Mockio.c] Error 1
```

**CI will not catch it.** `.github/workflows/unit_tests.yml:42` runs `make -C build`
serially, which passes cleanly (exit 0, 0 warnings). Only the developer loop is affected,
non-deterministically — and the exposure grows as other suites adopt `MOCK_HEADERS`.

The `CMOCK_MOCKED_HEADERS` global property is meant to generate each mock once, and it does
prevent duplicate `add_custom_command()` registrations — but it cannot stop the Makefile
generator replicating that single rule per target. Counting CMock invocations in a clean
`-j8` build shows the deduplication is already fully defeated: **18 invocations either way**,
with or without the guard. It saves nothing and is the reason the outputs are shared.

#### Proposed fix

Give each target its own CMock output directory and config file, and drop the guard. The
change is confined to `cmake/LedgerUT.cmake`, so it fixes every suite — present and future —
with no edit to any suite's `CMakeLists.txt`, no change to the `ledger_unit_tests_add_test()`
API, and no CMake version bump (3.10 is enough; generator-agnostic).

Ready-to-apply patch: [`ledgerut-cmock-race.patch`](./ledgerut-cmock-race.patch) (+23 / −26).
In `ledger_unit_tests_init()` the shared mock directory becomes a root and the single shared
`cmock_config.yml` disappears; in `ledger_unit_tests_add_test()` each target gets its own
`mocks/<target>/` directory and its own `cmock_config_<target>.yml`, and the
`CMOCK_MOCKED_HEADERS` property is deleted. A per-target config file is required because
`cmake/cmock_config.yml.in` hard-codes the output location as `:mock_path: '@CMOCK_GEN_DIR@'`.

Mocks then land in `build/mocks/<target>/Mock<header>.c`, so each output is owned by exactly
one target and no competing rule can exist.

Measured over 5 clean from-scratch `-j8` builds each:

| | Result | Wall clock |
|---|---|---|
| Baseline | **3 passed / 2 failed** | 4.1–4.3 s (passing runs) |
| Patched | **5 passed / 0 failed** | 2.4–3.4 s |

Serial build with the patch: exit 0, 0 warnings, `100% tests passed, 0 tests failed out of 4`,
coverage unchanged. It is faster despite the same 18 generations, because the invocations no
longer contend on shared paths.



The patch:  [ledgerut-cmock-race.patch.txt](https://github.com/user-attachments/files/30786391/ledgerut-cmock-race.patch.txt)

unit-tests/address_book/CMakeLists.txt
cedelavergne-ledger: Done, added more tests

fuzzing/sdk-fuzz/harness/fuzz_address_book.c
cedelavergne-ledger: Done, added more tests

iartemov-ledger (commented):
**2nd round.** 
2 more findings of a friend of mine.
The High one seems to be relevant. 
It's difficult to confirm the the Medium one.

fuzzing/sdk-fuzz/generate-seeds-address-book.py
cedelavergne-ledger: Fixed

unit-tests/address_book/CMakeLists.txt
cedelavergne-ledger: Let's postpone that to a dedicated PR....

fuzzing/sdk-fuzz/harness/fuzz_address_book.c
iartemov-ledger: The PR now ships **two libFuzzer fuzz targets** for the address book, where round 1 had one:

| Fuzz target | Entry point | Added |
|---|---|---|
| `fuzz_address_book` | `addr_book_handle_apdu()` — fuzzed APDUs through chunk reassembly, P1 dispatch and TLV parsing | round 1 |
| `fuzz_address_book_crypto` | the `address_book_crypto.c` API directly, bypassing TLV parsing | **new**, answers R1 #2 |

Both are declared in `SDK_FUZZ_TARGETS` (`fuzzing/sdk-fuzz/CMakeLists.txt`) and
`fuzz-manifest.toml`, and both were exercised in the campaign the figures come from — a clean
run with **0 crashes and 0 replay failures** on each (110 and 31 replay inputs). See §1.3 for
why earlier runs were unusable.

Union over the two fuzz targets, key files only:

| | Round 1 | Round 2 |
|---|---|---|
| **Regions** | 25.44 % | **42.92 %** (388/904) |
| **Lines** | 23.81 % | **40.42 %** (713/1764) |
| **Functions** | 42.11 % | **62.41 %** (83/133) |
| **Branches** | 20.34 % | **33.09 %** (135/408) |

Per file (regions / lines / branches), with the round-1 region figure for comparison:

| File | R1 regions | R2 regions | R2 lines | R2 branches |
|---|---|---|---|---|
| `address_book.c` | 83.67 % | **93.88 %** | 84.15 % | 85.71 % |
| `address_book_crypto.c` | 0.00 % | **90.18 %** | 89.86 % | 76.00 % |
| `address_book_common.c` | 74.29 % | 74.29 % | 65.15 % | 87.50 % |
| `ledger_account_provide_contact.c` | 40.32 % | 40.32 % | 41.38 % | 19.23 % |
| `ledger_account_edit.c` | 20.51 % | 37.18 % | 35.85 % | 20.59 % |
| `identity_edit_contact_name.c` | 16.44 % | 32.88 % | 27.63 % | 21.43 % |
| `ledger_account_register.c` | 14.52 % | 32.26 % | 32.00 % | 14.29 % |
| `identity_edit_identifier.c` | 20.54 % | 30.36 % | 28.05 % | 17.39 % |
| `identity_provide_contact.c` | 22.22 % | 27.78 % | 29.38 % | 13.89 % |
| `identity_register.c` | 34.75 % | 27.97 % | 26.14 % | 12.96 % |
| `identity_edit_scope.c` | 15.04 % | 22.12 % | 20.89 % | 10.42 % |

Two things stand out:

- **`address_book.c` 83.67 % → 93.88 % regions** (lines 75.61 % → 84.15 %, branches
  76.19 % → 85.71 %) — the P2 / two-chunk fix landing exactly where round 1 predicted.
- **`address_book_crypto.c` 0 % → 90.18 %**, 100 % of functions, entirely from the new
  direct harness. Through the TLV/APDU path (`fuzz_address_book` alone) it is *still* at
  **0.00 %** — the parsing bottleneck is unchanged, and §3.1 is why.

`identity_register.c` is the one regression (34.75 % → 27.97 %). Plausibly budget:
the TLV mutator grammar is Register-Identity-specific, and the two-chunk split now consumes
iterations that previously all went to single-chunk register payloads. Worth a longer run to
confirm rather than treating as significant.

*Note: the campaign's own `key-files-coverage.txt` prints a TOTAL of 48.13 % regions, but it
lists `address_book_crypto.c` twice (once per fuzz target) and so double-counts it. The union
figures above are de-duplicated.*

unit-tests/address_book/CMakeLists.txt
mbrousset-ledger: fixed in [9e13fb4](https://github.com/LedgerHQ/ledger-secure-sdk/pull/1666/commits/9e13fb48211ebc7db2918396cafd1d3e0b8433e9) in a different fashion, mock generation moved at configure time. It is simpler and we keep one unique and global mock

fuzzing/sdk-fuzz/harness/fuzz_address_book.c
cedelavergne-ledger: Reviewed and fixed

unit-tests/address_book/CMakeLists.txt
iartemov-ledger: https://github.com/LedgerHQ/ledger-secure-sdk/issues/1676

iartemov-ledger (approved):
Cool work!

PR #1673, opened 2026-08-06
fuzzing: simplify the app harness contract and consolidate the tooling

N3ur0sis:
## Description

An app harness now describes only what it dispatches, not how input is laid out.

**App contract.** Harnesses used to repeat a fixed preamble: `scenario_layout.h`,
`FUZZ_PREFIX_SIZE_FALLBACK`, `FUZZ_CTRL_OFF`/`FUZZ_CTRL_LEN`, a
`fuzz_lane_is_structured()` macro, an `LLVMFuzzerCustomMutator()` that only forwarded,
a hand-written `fuzz_n_commands`, an empty `fuzz_app_reset()`, and a `fuzz_entry()`
that only called `fuzz_harness_entry()`. All defaulted in the framework now;
`fuzz_layout_check.h` and `scenario_layout.h` are deleted. What remains is
`fuzz_commands[]`, `FUZZ_COMMAND_COUNT()` and `fuzz_app_dispatch()`.

**Prefix arithmetic.** Harnesses with their own grammar no longer read
`absolution_globals_size`, bounds-check it and slice the tail by hand — that is one
`fuzz_mutate_input_with()` call. The prefix size is emitted at configure time
(`cmake/EmitPrefixSize.cmake`), so no app declares it. `fuzz_base58.c` 82 → 60 lines,
`fuzz_tlv_trusted_name.c` 99 → 66.

**Shared mocks.** The PKI gate override, previously duplicated in three trees, is
promoted to `fuzzing/mock/pki/ledger_pki_policy.c` and is now fuzzer-controlled rather
than hardcoded success. `fuzz_bip32.h` / `fuzz_bytes.h` factor out harness helpers.

**Tooling.** `fuzzing/scripts/` 13 → 8 files: seven single-purpose scripts collapse
into `invariant.py` (sync and tune always ran together) and `seeds.py`.

**Docs.** Reference pages follow the contract change. `ci.dox` also documents the CI
corpus lifecycle and what to do when Codecov reports a coverage drop.

**CI.** One line, `secrets: inherit`, so the reusable workflow's coverage job can reach
`CODECOV_TOKEN` (LedgerHQ/ledger-app-workflows#279).

## Changes include

- [ ] Tests
- [X] Documentation
- [X] Other (for changes that might not fit in any category)

## Breaking changes

The framework now provides `fuzz_entry()` and `LLVMFuzzerCustomMutator()` as strong
definitions, so a harness defining its own hits a duplicate symbol. Either delete the
forwarder, or opt out before including `fuzz_harness.h` with `FUZZ_APP_CUSTOM_ENTRY`,
`FUZZ_APP_CUSTOM_MUTATOR`, or `FUZZ_APP_POST_MUTATE`.


.github/workflows/clusterfuzzlite.yml
```diff
@@ -16,3 +16,5 @@ jobs:
     uses: LedgerHQ/ledger-app-workflows/.github/workflows/reusable_clusterfuzz_tests.yml@v1
     with:
       exec_mode: ${{ github.event_name }}
+    # CODECOV_TOKEN, for the fuzzing coverage report.
+    secrets: inherit
```
cedelavergne-ledger: Maybe it is better to declare the expected secretw using `with` directive?

fuzzing/cmake/LedgerAppFuzz.cmake
```diff
@@ -144,6 +163,43 @@ unset(_LEDGER_FUZZ_DEFAULT_SANITIZERS)
...
+    # are stubbed), which hides everything behind it. mock/pki/ledger_pki_policy.c
+    # decides the verdict from the fuzzable fuzz_mock_pki_fail instead.
+    "-Wl,--wrap=check_signature_with_pki"
+    # Crypto the fuzzer cannot satisfy: real signing/keygen would reject every
+    # mocked value early, so these are intercepted and their verdict driven by
+    # fuzz_mock_crypto_fail. Bodies in mock/cx/cx_crypto.c.
+    "-Wl,--wrap=bip32_derive_with_seed_init_privkey_256"
+    "-Wl,--wrap=cx_ecfp_generate_pair_no_throw"
+    "-Wl,--wrap=cx_ecdsa_sign_no_throw"
+    "-Wl,--wrap=cx_ecschnorr_sign_no_throw")
```
cedelavergne-ledger, line 202: Why only those functions from lib_cxng?
Any possibility to have more global mocks for Apps, and validated SDK sources with the fuzzing?

fuzzing/doc/running.dox
```diff
@@ -46,13 +46,24 @@ Each run writes to `<app>/.fuzz-artifacts/<run-name>/`:
...
 
+@warning Zero crashes does not mean zero findings. UndefinedBehaviorSanitizer
+diagnostics are *recoverable* by default: the program prints `runtime error: ...`
+and carries on, so libFuzzer never treats the input as a crash and writes no
+artifact. The campaign therefore scans the worker logs and prints every reporting
+site, also saving them to `sanitizer-reports.txt`. **Read that file** — a run can
+exit 0 while having reported real integer truncations or overflows. Unsigned
+overflow specifically is compiled with `-fsanitize-recover=unsigned-integer-overflow`
+because deliberate wrapping is common (PRNGs, hashing), so it will always appear
+here rather than as a crash.
```
cedelavergne-ledger, line 65: Oh, this is important!
Do you think it can be handle by the generic reusable workflow, at least to write a specific message in the GITHUB_OUTPUT -or Annotations), and maybe produce an exit(1)?

fuzzing/include/fuzz_bytes.h
cedelavergne-ledger, line 1: Why not using macros from `os_utils.h` directly?

fuzzing/include/fuzz_harness.h
```diff
@@ -5,31 +5,56 @@
...
 #include "fuzz_defs.h"
+/* Optional: an app includes its own mocks.h only if it genuinely has
+ * app-specific stubs. The framework symbols this harness needs are declared
+ * in fuzz_defs.h. */
+#if defined(__has_include)
+#if __has_include("mocks.h")
 #include "mocks.h"
+#endif
+#else
+#include "mocks.h"
```
cedelavergne-ledger: I don't catch the difference with above lines and the interest to use `__has_include`?

fuzzing/include/fuzz_mutator.h
```diff
@@ -3,243 +3,176 @@
...
 
-#if FUZZ_APP_DATA_LEN > 0
-enum {
-    FUZZ_MUT_STRUCTURED_APP_DATA_CUTOFF    = 40,
-    FUZZ_MUT_STRUCTURED_POST_PREFIX_CUTOFF = 70,
-};
-#else
 enum {
-    FUZZ_MUT_STRUCTURED_POST_PREFIX_CUTOFF = 60,
+    FUZZ_MUT_INITIAL_INPUT_SIZE = 256,  ///< Size of the input synthesised for an empty seed.
```
cedelavergne-ledger: Why an enum here, and not simple `define`?
Usually enum is for organized list ?

fuzzing/include/fuzz_mutator.h
```diff
@@ -3,243 +3,176 @@
...
-
-static size_t fuzz_bootstrap_input(uint8_t *data, size_t size, size_t prefix_size, size_t max_size)
+/**
+ * @brief Where Absolution's sampled state ends.
+ *
+ * Read from @ref fuzz_absolution_prefix_size, which the build generates and links
+ * in. Never zero: EmitPrefixSize.cmake fails the build instead, since zero would
+ * degrade this to flat-byte mutation over structured state.
+ */
+static inline size_t fuzz_prefix_size(void)
```
cedelavergne-ledger: Why do you define the different functions here as `inline` (so duplicated in each source file including this header), rather than creating the correspondinf c file?

fuzzing/mock/fuzz_runtime.c
```diff
@@ -0,0 +1,94 @@
...
+void __wrap_os_explicit_zero_BSS_segment(void) {}
+
+/**
+ * @brief explicit_bzero routed to a volatile store loop.
+ *
+ * MSan cannot see through the SDK's version, so a SANITIZER=memory campaign
+ * reports use-of-uninitialised-value on every buffer it zeroed
+ * (google/sanitizers#1507). The barrier keeps the store from being elided.
+ */
+void *__wrap_explicit_bzero(void *dst, size_t len);
```
cedelavergne-ledger: Why do you need to declare it just before the implementation?

fuzzing/mock/tlv_mutator.c
cedelavergne-ledger, line 1: No possibility to directly use functions from `tlv_library`?

.github/workflows/clusterfuzzlite.yml
N3ur0sis: Good call,done. it now declares just codecov_token. It's required: false on the reusable side, so forks without the token still run, the upload step just skips.

fuzzing/cmake/LedgerAppFuzz.cmake
N3ur0sis: Only the functions that cannot run on a host are wrapped: seed derivation, key generation, ECDSA/Schnorr signing, and the PKI gate. Everything else in lib_cxng — hashing, AES, the math layer  ,runs for real, which is the point: those are the parts we want the fuzzer exercising rather than a stub. Wrapping more would trade real coverage for reachability we don't need, since the wrapped ones are only there because they'd reject every mocked value at the first check and hide everything behind them.

fuzzing/include/fuzz_bytes.h
N3ur0sis: You're right,  os_utils.h already has U2LE/U4LE/U8LE, U2BE/U4BE/U8BE and the matching U*_ENCODE. All six helpers were duplicates. i deleted fuzz_bytes.h and fuzz_bip32.h uses os_utils.h; i'll update the app's branch the same way so the two stay in step.

fuzzing/include/fuzz_mutator.h
N3ur0sis: Switched to #define. Fair point.

fuzzing/mock/fuzz_runtime.c
N3ur0sis: You're right, it was residual,  removed. -Wall -Wextra don't imply -Wmissing-prototypes, and the sibling __wrap_os_explicit_zero_BSS_segment in the same file never had one, so it was inconsistent as well as unnecessary.

fuzzing/include/fuzz_harness.h
N3ur0sis: Now just #if __has_include("mocks.h"). The conditional itself has to stay: an app only ships mock/mocks.h if it has app-specific stubs, and some have none,app-boilerplate ships no such file so an unconditional include wouldn't compile there?

fuzzing/include/fuzz_mutator.h
N3ur0sis: Dropped it and moved the bodies into fuzzing/mock/fuzz_mutator.c; the header keeps declarations, and the three internal helpers are now static in the .c.

fuzzing/mock/tlv_mutator.c
N3ur0sis: lib_tlv is the parser, handler callbacks, get_*_from_tlv_data(), tlv_check_received_tags(), and exposes no encoder. tlv_mutator.c goes the other way: it emits TLV bytes to feed that parser.

cedelavergne-ledger (approved):

PR #1709, opened 2026-09-21
io: arm the TOCTOU latch for deferred default-CLA commands

cedelavergne-ledger:
## Description

A default-CLA command that returns SWO_NO_RESPONSE left the latch unarmed, so a second command was accepted while its review was still on screen.

## Changes include

- [ ] Tests
- [ ] Documentation


io_legacy/src/os_io_legacy.c
```diff
@@ -434,6 +434,14 @@ int io_legacy_apdu_rx(uint8_t handle_ux_events)
                         buffer_out_length = 0;
                     }
                     if (err == SWO_NO_RESPONSE) {
+                        // TOCTOU: the command deferred its reply, typically because it put
```
iartemov-ledger: Could you simplify and shorten up the comment?

io_legacy/src/os_io_legacy.c
```diff
@@ -434,6 +434,14 @@ int io_legacy_apdu_rx(uint8_t handle_ux_events)
...
+                        // a review on screen. Arm the latch here too, otherwise a command
+                        // handled in this branch is the one case left unprotected: a second
+                        // command would be accepted while the review is still displayed and
+                        // could overwrite the state that review is about to apply.
+                        // io_legacy_apdu_tx() releases the latch when the reply is finally
+                        // sent, on the accept, reject and error paths alike.
+                        os_io_set_reply_pending(true);
```
iartemov-ledger, line 439: Is this for only Address Book feature today (the one that uses `SWO_NO_RESPONSE`)?
Here we call app's entry point and return `SWO_NO_RESPONSE`: https://github.com/LedgerHQ/ledger-secure-sdk/blob/master/app_features/address_book/src/ledger_account_register.c#L307. 
It seems that there is a place in `app-ethereum` (the only Address Book feature consumer) where we may return without calling the `review_choice` callback: https://github.com/LedgerHQ/app-ethereum/blob/develop/src/features/address_book/handle_ledger_account.c#L168.
In this case SDK's ` io_send*()` is not called and the latch is never reset. 

The fix would be in `app-ethereum` to add `choice_callback(false);` in the branch.

io_legacy/src/os_io_legacy.c
cedelavergne-ledger: Sure. Done

io_legacy/src/os_io_legacy.c
cedelavergne-ledger: This feature will be also available in other Apps, so the good fix is really here.
Besides, it could also be used when we'll develop a next common feature directly in the SDK

iartemov-ledger: I was talking about an existing vulnerability in `app-ethereum` that has become worse with the SDK fix.
I'm not talking about a moving the entire fix to `app-ethereum`.
So as 2nd step after this PR is merged is to apply the proposed fix above (`choice_callback(false);`) to `app-ethereum`.

iartemov-ledger (approved):

